PR #8, opened 2024-10-09
A first Lie group!

kellertuer:
This PR aims to test our interface ideas with a first concrete Group operation and a Lie group

* [x] Introduce `AdditionGroupOperation`
* [x] Introduce `TranslationGroup`
* transfer default for the addition group (and test snippets)
  * [x] `exp`
  * [x] `log`
  * [x] `diff_conjugate`
  * [x] `div_left_compose`, `diff_right_compose`
  * [x] `diff_inv`
  * [x] `lie_bracket`
  * [x] document them thoroughly to help people getting started what the terminology is 
* transfer in general 
  * [x] `is_point` and for identity `check_vector` (a bit nicer than before)
  * [x] `is_vector`
* start a test suite
  * add tests for the translation group
  * add generic tests with a dummy Lie group and a dummy Lie algebra to test a few fallbacks.
  * [x] tests for also the default given ones like `inv_left_compose`, `adjoint`, `inv_right_compose`
* [x] reorganise the interface to have a bit of a shorter page – maybe a structure with sub pages for
   * general interface functions
   * group operations interface
   * group actions interface
* [x] check for ambiguities once the `Identity` gets involved 
* [x] setup codecov
* [x] setup vale (though not yet run on CI)
* [x] setup/prepare zenodo metadata.


kellertuer:
@mateuszbaran I struggle a bit with the next const-type-thing. This time LieAlgebra as a special type of Fiber.

```
using LieGroups
G = TranslationGroup(3)
LieAlgebra(G)
```

yields

```
TypeError: in Fiber, in G, expected G<:(LieGroup{ℝ, O, M} where {O<:AbstractGroupOperation, M<:ManifoldsBase.AbstractManifold{ℝ}}), got a value of type TranslationGroup{ℝ, ManifoldsBase.TypeParameter{Tuple{3}}}
Stacktrace:
 [1] (ManifoldsBase.Fiber{𝔽, ManifoldsBase.TangentSpaceType, G, I} where {𝔽, G<:(LieGroup{𝔽, O, M} where {O<:AbstractGroupOperation, M<:ManifoldsBase.AbstractManifold{𝔽}}), I<:Identity})(G::TranslationGroup{ℝ, ManifoldsBase.TypeParameter{Tuple{3}}})
   @ LieGroups ~/Repositories/Julia/LieGroups.jl/src/interface.jl:91
 [2] top-level scope
   @ REPL[4]:1
```

I am super buzzed by “expected Lie Group got TranslationGroup”, because TranslationGroup _is_ a LieGroup (const)

besides that this PR only misses to check which other functions already to take over (from [here](https://github.com/JuliaManifolds/Manifolds.jl/blob/master/src/groups/addition_operation.jl)). and maybe work a bit on the test suite further.

Besides fighting parameters in const (constantly it seems) – the interface seems usable I think, so we could start transferring more LieGroups afterwards.

mateuszbaran:
It appears you just forgot one `typeof` :slightly_smiling_face:

kellertuer:
Oh noes! I was too busy checking the third parameter over and over again, that I did not spot that one – and probably got confused that the _variable_ has a capital letter. Thanks!

kellertuer:
I am also fond of getting the test suite better from the start here – slowly doing that along with the first group. If that works well I will adopt the same scheme over in Manifolds.jl

src/group_operations/addition.jl
```diff
@@ -0,0 +1,81 @@
...
+end
+function is_identity(
+    G::LieGroup{𝔽,AdditionGroupOperation}, h::Identity{AdditionGroupOperation}; kwargs...
+) where {𝔽}
+    return true
+end
+function is_identity(
+    G::LieGroup{𝔽,AdditionGroupOperation}, h::Identity; kwargs...
+) where {𝔽}
+    return true
```
mateuszbaran: Shouldn't this be `false`? If `h` is addition identity, the method above should catch it and for non-addition identities I'd expect `false` here.

kellertuer: Yes, correct. I hit one of these logical things just now in exp/log fallbacks, and will work through the interface functions while writing the test functions.
This could probably also already be checked on the interface level and no longer here (same Op-> true, different -> false)

mateuszbaran:
You're welcome.

I'm still curious how this testing approach works out but I guess I'll wait until it's a bit more complete.

src/group_operations/addition.jl
mateuszbaran: Yes, this check can be interface-level for sure.

kellertuer:
If you have comments / questions feel free to post them, the general idea is:

With a bit of given data of (up to) 3 points, 3 vectors

* provide modular test functions for every single function of the interface – potentially with interactions to others like in exp/log
* provide `properties` that list which functions or more general which things can be tested
* provide `expectations` for whenever we have to be able to know a result

For now both `properties` and `expectations` are set manually, but I started already with an idea in the Manifolds.jl branch to have an “auto-discover” mode for them.

This should make testing concrete other groups later a bit faster.
I also tried to encapsulate all this in a small test package. so that this can easily be loaded and avoid the need for the “if not yet loaded include now” approach we have in a few packages for generic tools like a `DummyLieGroup` for example.

The interface/fallbacks in general have to be done as a separate test anyways.

src/group_operations/addition.jl
kellertuer: Fixed it.

mateuszbaran:
> * provide `expectations` for whenever we have to be able to know a result

Could you add specific expectations to tested functions in the `TranslationGroup` test suite? For example that `exp` performs addition and `log` does subtraction?

> For now both `properties` and `expectations` are set manually, but I started already with an idea in the Manifolds.jl branch to have an “auto-discover” mode for them.

I think auto-discovery shouldn't be that important because when you develop some code, you know what should work. So it's not much work to add it manually. Maybe a little bit of semi-automatic discovery, such as "if you declared X and Y as implemented, Z should also work", would be useful though.

kellertuer:
> > * provide `expectations` for whenever we have to be able to know a result
> 
> Could you add specific expectations to tested functions in the `TranslationGroup` test suite? For example that `exp` performs addition and `log` does subtraction?

Uff. No clue, that is maybe a bit too high-level.
For now `exp` and `log` are mainly checked to produce valid points and vectors, respectively and if both are available, that they are inverses of each others for the given points.

The expectations I have in mind is mainly numbers and strings – main example for now the string `repr(G)` returns.
You idea sounds much more like a whole AI to implement to interprete results. That is not what expectation means – much simpler, just in the sense of expected values (for certain things).

> 
> > For now both `properties` and `expectations` are set manually, but I started already with an idea in the Manifolds.jl branch to have an “auto-discover” mode for them.
> 
> I think auto-discovery shouldn't be that important because when you develop some code, you know what should work. So it's not much work to add it manually. Maybe a little bit of semi-automatic discovery, such as "if you declared X and Y as implemented, Z should also work", would be useful though.

Sure auto.discover is something not really urgent and also a bit dangerous, since you would irregularly run it and it would save both properties and expectations in a `.jld2` file. The danger is that it might also save erroneous expectatios then (in the sense of wrong values/strings). I would also not consider this too high priority, it was just something the manifolds-test suite branch experimented with a bit. For now we should maybe not follow this here.

mateuszbaran:
> Uff. No clue, that is maybe a bit too high-level. For now `exp` and `log` are mainly checked to produce valid points and vectors, respectively and if both are available, that they are inverses of each others for the given points.
> 
> The expectations I have in mind is mainly numbers and strings – main example for now the string `repr(G)` returns. You idea sounds much more like a whole AI to implement to interprete results. That is not what expectation means – much simpler, just in the sense of expected values (for certain things).
> 

That's fine but then this test suite is not sufficiently exhaustive to replace other tests. And I definitely don't mean anything related to AI, only letting the user provide reference results for specific operations such as exp and log. Maybe it's obvious but I wouldn't like to have an expectation that when `test_LieGroup(G, properties, expectations)` passes, the group is definitely correctly implemented. Similarly `test_manifold` only checks some general properties and is not exhaustive.

kellertuer:
The suite is meant to provide a framework for common, usual tests. Sure that might not be exhaustive.
I would even be open, that _none_ of our tests are probably exhaustive (in favour of a finite time).

But the point that the suite is to provide tests for most common properties to check is probably a good thing to add to their docs.
Though I think we might be able to cover as much as we usually test in Manifolds.jl for most manifolds (and not only with the test_manifold function, but a bit beyond).

src/groups/translation_group.jl
```diff
@@ -0,0 +1,23 @@
...
+"""
+const TranslationGroup{𝔽,T} = LieGroup{𝔽,AdditionGroupOperation,Manifolds.Euclidean{T,𝔽}}
+
+function TranslationGroup(n::Int...; kwargs...)
+    Rn = Manifolds.Euclidean(n...; kwargs...)
+    return TranslationGroup{typeof(Rn).parameters[[2, 1]]...}(Rn, AdditionGroupOperation())
+end
+
+function Base.show(io::IO, G::TranslationGroup)
+    return print(io, "TranslationGroup($(G.manifold), $(G.op))")
```
mateuszbaran: I guess `G.op` shouldn't be necessary, and we should have a matching constructor from `G.manifold`??

src/groups/translation_group.jl
```diff
@@ -0,0 +1,23 @@
...
+    TranslationGroup{𝔽,T}
+
+The Lie group consisting of the [`AdditionGroupOperation`](@ref) on some
+[`Euclidean`](@extref `Manifolds.Euclidean`) space.
+
+# Constructor
+    TranslationGroup(n₁,...,nᵢ; kwargs...)
+
+Generate the translation group on ``𝔽^{n₁,…,nᵢ}`` = `Euclidean(n₁,...,nᵢ; field=𝔽)`,
+which is isomorphic to the group itself. all keyword arguments in `kwargs...`
```
mateuszbaran:
```suggestion
which is isomorphic to the group itself. All keyword arguments in `kwargs...`
```

test/LieGroupsTestSuite.jl/LieGroupsTestSuite.jl
```diff
@@ -0,0 +1,208 @@
...
+
+This module provides tools and dummy structures to test functionality provided
+within `LieGroups.jl`.
+
+For every test function, several interactions to other functions can be activated.
+The following functions are expected to be available, since their defaults just pass through to the manifold
+* `is_point` both on the Lie group `G` and the Lie algebra `𝔤`
+* `isapprox(G,g,h)` and `issaprox(𝔤, X, Y)`
+* `copy(G,g)`
+* `norm(𝔤,X)`
```
mateuszbaran:
```suggestion
* `copy(G, g)`
* `norm(𝔤, X)`
```

test/LieGroupsTestSuite.jl/LieGroupsTestSuite.jl
```diff
@@ -0,0 +1,208 @@
...
+end
+
+"""
+    test_exp_log(G, g, h, X)
+
+Test functionality of `exp` and `log` for given Lie group elements `g`, `h` and
+a vector `X` from the Lie Algebra.
+
+!!! note
+    This function requires`Is_point(G, g)` and `is_point(𝔤, X)` to be implemented
```
mateuszbaran:
```suggestion
    This function requires `is_point(G, g)` and `is_point(LieAlgebra(G), X)` to be implemented
```
Maybe that would be more clear?

src/groups/translation_group.jl
kellertuer: well, spotted, we indeed have one that just requires the dimensions of the manifold even. Will check how to update this.

test/LieGroupsTestSuite.jl/LieGroupsTestSuite.jl
kellertuer: sure.

src/groups/translation_group.jl
kellertuer: Was even easy to fix. Thanks for spotting, the show is now much nicer with `TranslationGroup(3; field=...)`

kellertuer:
I think I at least see an end to this with only a few small test suite functions and a few lines of code implementation left.

olivierverdier:
Speaking about testing, I wrote a [package containing sensible tests for Lie groups](https://olivierverdier.github.io/ManifoldGroupTesting.jl/dev/#Group-Test-Helpers). You can see an example of usage here: https://github.com/olivierverdier/ManifoldGroupTesting.jl/blob/main/test/test_group.jl
There are also tests for group actions.

kellertuer:
Nice!
Here for now I just wrote a few functions like that in an internal sub-package, though here the blocks in each functions are actually already test sets, so they usually test a few more things than just one thing – for now they are a bit more focussed on low-level correctness – that the code works, yours already seem to also check the next level, that more formulae are verified.

src/interface.jl
```diff
@@ -437,24 +498,114 @@ function inv! end
...
+The first variant calls [`is_point`](@extref ManifoldsBase.is_point) on the [`LieAlgebra`](@ref) `𝔤` of `G`.
+The second variant calls [`is_vector`](@extref ManifoldsBase.is_vector) on the $(_link(:AbstractManifold)) at the [`identity_element`](@ref).
+
+All kwyword arguments are passed on to the corresponding call
+"""
+
+@doc "$(_doc_is_vector)"
+ManifoldsBase.is_vector(G::LieGroup, X; kwargs...) = ManifoldsBase.is_point(
+    LieAlgebra(G), X; kwargs...
+)
```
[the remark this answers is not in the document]
kellertuer: Hmpf. Not sure where this's from, locally I do not have this.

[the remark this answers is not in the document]
kellertuer: Formatter changed this a bit a go and now back. No clue why.

kellertuer:
In theory everything is documented by now.

In practice I fought “duplicate docs” errors for the last hour, that I have never seen before coming from one and the same autodocs block. Well. Maybe this magically disappears tomorrow, or someone else sees how to fix this.

Besides that this is finished.

kellertuer:
Hmpf. Who ever came up with this clever idea in `@autodocs`

```
Note that page matching is done using the end of the provided strings and so a.jl will be matched by any source file that ends in a.jl, i.e. src/a.jl or src/foo/a.jl.
```

never wrote two interfaces in one project.

docs/src/groups/translation.md
```diff
@@ -0,0 +1,7 @@
+# The Lie group of translations on an Euclidean space
```
mateuszbaran:
```suggestion
# The Lie group of translations on a Euclidean space
```

docs/src/interface/operations.md
```diff
@@ -0,0 +1,17 @@
+
+# An interface for Lie group operations
+
+```@docs
+AbstractGroupOperation
+Identity
+```
+
+For some generic group operations, generic implementations are available.
```
mateuszbaran:
```suggestion
For some generic group operations, a generic implementation is available.
```
AFAIK the plural of "implementation" is rarely used.

docs/src/notation.md
```diff
@@ -5,16 +5,14 @@ For most of the notation used throughout this package, we follow [HilgertNeeb:20
...
-| ``λ_g: \mathcal G → \mathcal G`` | the left multiplication map ``λ_g(h) = g∘h`` | |
+| ``λ_g: \mathcal G → \mathcal G`` | the left group operation map ``λ_g(h) = g∘h`` | |
 | ``\mathcal M`` | a manifold | ``\mathcal M_1, \mathcal M_2,\ldots,\mathcal N`` | |
-| ``\operatorname{Exp}`` | the matrix exponential | |
-| ``\operatorname{Log}`` | the matrix logarithm | |
-| ``ρ_g: \mathcal G → \mathcal G`` | the right multiplication map ``ρ_g(h) = h∘g`` | |
-| ``σ: \mathcal G × \mathcal M`` | a left group action | we use ``σ_g(p)`` if the group element is fixed |
-| ``τ: \mathcal M × \mathcal G`` | a right group action | ``σ_\mathrm{R}`` | we use ``τ_g(p)`` if the group element is fixed |
+| ``ρ_g: \mathcal G → \mathcal G`` | the right group operation map ``ρ_g(h) = h∘g`` | |
+| ``σ: \mathcal G × \mathcal M`` | a left group action | ``σ_g(p)`` to emphasize a fixed grou pelement |
```
mateuszbaran:
```suggestion
| ``σ: \mathcal G × \mathcal M`` | a left group action | ``σ_g(p)`` to emphasize a fixed group element |
```

docs/src/notation.md
```diff
@@ -5,16 +5,14 @@ For most of the notation used throughout this package, we follow [HilgertNeeb:20
...
+| ``τ: \mathcal M × \mathcal G`` | a right group action | ``σ_\mathrm{R}`` | ``τ_g(p)`` o emphasize a fixed grou pelement |
```
mateuszbaran:
```suggestion
| ``τ: \mathcal M × \mathcal G`` | a right group action | ``σ_\mathrm{R}`` | ``τ_g(p)`` to emphasize a fixed group element |
```

docs/src/references.md
```diff
@@ -1,7 +1,7 @@
 # Literature
 
 This is all literature mentioned / referenced in the `LieGroups.jl` documentation.
-Usually you find a small reference section at the end of every documentation page
+You find a small reference section at the end of every documentation page
```
mateuszbaran:
```suggestion
You can find a small reference section at the end of every documentation page
```

src/LieGroups.jl
```diff
@@ -1,38 +1,48 @@
...
+export TranslationGroup
 
-export apply, apply!, diff_apply, diff_apply!
-export base_manifold
-export adjoint,
-    adjoint!,
-    compose,
-    compose!,
+export adjoint, adjoint!, apply, apply!
+export base_Lie_group, base_manifold
```
mateuszbaran: Can we stop using capital letters in function names? They make them look inconsistent.

src/Lie_algebra/Lie_algebra_interface.jl
```diff
@@ -0,0 +1,86 @@
...
+
+# Constructor
+
+    LieAlgebra(G::LieGroup)
+
+Return the Lie Algebra belonging to the [`LieGroup`](@ref) `G`.
+"""
+const LieAlgebra{𝔽,G,I} = ManifoldsBase.Fiber{
+    𝔽,ManifoldsBase.TangentSpaceType,G,I
+} where {𝔽,G<:LieGroup{𝔽},I<:Identity}
```
mateuszbaran:
```suggestion
const LieAlgebra{𝔽,G} = ManifoldsBase.Fiber{
    𝔽,ManifoldsBase.TangentSpaceType,G,Identity{O}
} where {𝔽,O<:AbstractGroupOperation,G<:LieGroup{𝔽,O}}
```
This should simplify things a bit.

src/Lie_algebra/Lie_algebra_interface.jl
```diff
@@ -0,0 +1,86 @@
...
+    return ManifoldsBase.is_vector(G.manifold, e, X; kwargs...)
+end
+
+# Move this line already to ManifoldsBase? On Fibers of course.
+LinearAlgebra.norm(𝔤::LieAlgebra, X) = LinearAlgebra.norm(𝔤.manifold, 𝔤.point, X)
+function LinearAlgebra.norm(
+    G::LieGroup{𝔽,O}, ::Identity{O}, X
+) where {𝔽,O<:AbstractGroupOperation}
+    return LinearAlgebra.norm(G, identity_element(G), X)
+end
```
mateuszbaran: I'd expect `norm` for `VectorSpaceFiber` in `ManifoldsBase.jl` to already handle this.

src/group_actions/group_operation_action.jl
```diff
@@ -172,6 +208,10 @@ Return the inverse of the [`InverseRightGroupOperation`](@ref), that is the [`Ri
 """
 Base.inv(::InverseRightGroupOperation) = RightGroupOperation()
 
+function Base.show(io::IO, A::GroupOperationAction)
+    return print(io, "GroupOperationAction($(A.type),$(A.group))")
```
mateuszbaran:
```suggestion
    return print(io, "GroupOperationAction($(A.type), $(A.group))")
```

src/interface.jl
```diff
@@ -54,82 +52,18 @@ See also [`identity_element`](@ref) on how to obtain the corresponding [`Abstrac
 
 # Constructors
 
-    Identity(::LieGroup{𝔽,M,O}) where {𝔽,M,O<:AbstractGroupOperation}
+    Identity(::LieGroup{𝔽,O}) where {𝔽,M,O<:AbstractGroupOperation}
```
```suggestion
    Identity(::LieGroup{𝔽,O}) where {𝔽,O<:AbstractGroupOperation}
```

src/interface.jl
```diff
@@ -160,116 +94,99 @@ end
...
+    return ManifoldsBase.check_point(G.manifold, g; kwargs...)
+end
+function ManifoldsBase.check_point(
+    G::LieGroup{𝔽,O}, e::Identity{O}; kwargs...
+) where {𝔽,O<:AbstractGroupOperation}
+    return nothing
+end
+function ManifoldsBase.check_point(
+    G::LieGroup{𝔽,O}, e::Identity{O2}; kwargs...
+) where {𝔽,O<:AbstractGroupOperation,O2}
```
```suggestion
) where {𝔽,O<:AbstractGroupOperation,O2<:AbstractGroupOperation}
```

mateuszbaran (commented):
A few comments. I haven't checked everything yet.

docs/src/interface/operations.md
kellertuer: what would you prefer than for these generic function definitions?

src/LieGroups.jl
kellertuer: for `Lie` I was just not sure, since it is a last name, I would always capitalise that, for example also in adjectives. But I also do not have a strong opinion, just my personal taste would be to keep `_Lie_` in function names. If you really do not like that, we can surely rename that.

src/Lie_algebra/Lie_algebra_interface.jl
kellertuer: `Identity` does not exist in `ManifoldsBase.jl`, only in `Manifolds.jl`.
We would also remove `Identity` from `Manifolds.jl`.

kellertuer: Or do you mean line 69? No that is not yet handled, but it would be nice to have that in ManifoldsBase.

src/Lie_algebra/Lie_algebra_interface.jl
kellertuer: Hm, I think this broke something. I'd have to check.

_edit:_ For now I undid this change, since I do not understand how your fix could be made working.

src/Lie_algebra/Lie_algebra_interface.jl
mateuszbaran: I'm not sure about the method in line 70 but at least the one in line 69 should be handled by https://github.com/JuliaManifolds/ManifoldsBase.jl/blob/5972cde434c221009f354a229f7009856d4b0747/src/VectorFiber.jl#L11 .

docs/src/interface/operations.md
mateuszbaran: This comment doesn't refer to the code, only docstring and grammar.

src/Lie_algebra/Lie_algebra_interface.jl
kellertuer: No. line 69 does not have an “outer `p`”. It just reads `norm(𝔤,X)` (not `norm(𝔤,p,X)`)

docs/src/interface/operations.md
kellertuer: My question as well ;) I am really just not sure what to write instead.

src/Lie_algebra/Lie_algebra_interface.jl
mateuszbaran: Right, I didn't notice. Anyway, the method in line 69 could probably be introduced already in ManifoldsBase.jl for any `VectorSpaceFiber`?

kellertuer: That is what line 68 is asking already, sure we could do that. _Then_ this line is really obsolete and we just need the l.70 for completeness.

src/LieGroups.jl
mateuszbaran: Thanks. I prefer when function names don't use capital letters at all.

kellertuer: Sure. I am just German, wE cApItAliZe a lot. But as long as it is consistent, I do not mind – hope to remember then (because German).

src/Lie_algebra/Lie_algebra_interface.jl
mateuszbaran: I've figured out a slightly different approach and committed it.

docs/src/interface/operations.md
kellertuer: Rephrased it to “Some specific group operations allow to define functions for all manifolds they
do build a Lie group with. These are”

starting an overview afterwards, because this will list the general multiplication and the matrix multiplication case and which ever one we come up with generically.

Affie:
Hi, I didn't subscribe to LieGroups.jl notifications after it was transferred and now I'm far behind so sorry I can't help out here. It's very nice to see that you made a lot of progress!

kellertuer:
No worries, hat was quite a technical hustle we went through, but (yay!) we got the name that way. Feel free to comment anything you notice or open issues about groups you would like to see. We can also already have your group in mind you mentioned, just open an issue for that as well.

src/Lie_algebra/Lie_algebra_interface.jl
mateuszbaran: I've made a PR to ManifoldsBase.jl with that method.

kellertuer: Nice. I noticed that we might not run Aqua here yet, I should add that!

kellertuer:
I started looking a bit into ambiguities, and in general it looks nice, just that for example

```
+(Manifolds.Identity{O}, LieGroups.Identity{O}) 
```

is a bit of a temporary and annoying one. I will put that one on Ignore since it will resolve automatically once we make the breaking change in Manifolds.

kellertuer:
So we now further
* [x] run `Aqua.jl` tests
* [x] the test suite can now run less tests (skip mutating variants) by setting `:Mutating => false` in the properties, either to be a bit faster or to work on a manifold with non-mutating points/vectors.

docs/src/interface/operations.md
```diff
@@ -0,0 +1,20 @@
+
+# An interface for Lie group operations
+
+```@docs
+AbstractGroupOperation
+Identity
+```
+
+Some specific group operations allow to define functions for all manifolds they
+do build a Lie group with. These are
```
mateuszbaran: I think you mean functions for Lie groups and not manifolds?

src/Lie_algebra/Lie_algebra_interface.jl
```diff
@@ -0,0 +1,86 @@
+
+"""
+    LieAlgebra{𝔽, G} <: AbstractManifold{𝔽}
+
+Represent the Lie Algebra ``$(_math(:𝔤))``, that is a ``𝔽``vector space with an associated
```
mateuszbaran:
```suggestion
Represent the Lie algebra ``$(_math(:𝔤))``, that is a ``𝔽`` vector space with an associated
```

src/Lie_algebra/Lie_algebra_interface.jl
```diff
@@ -0,0 +1,86 @@
...
+[`lie_bracket`](@ref) ``[⋅,⋅]: $(_math(:𝔤))×$(_math(:𝔤)) → $(_math(:𝔤))`` which fulfills
+
+1. ``[X,X] = 0`` for all ``X ∈ $(_math(:𝔤))``
+2. The Jacobi identity holds ``[X, [Y,Z]] = [[X,Y],Z] = [Y, [X,Z]]`` holds for all ``X, Y, Z ∈ $(_math(:𝔤))``.
```
mateuszbaran: "holds" appears twice here

src/Lie_algebra/Lie_algebra_interface.jl
```diff
@@ -0,0 +1,86 @@
...
+
+The Lie algebras considered here are those related to a [`LieGroup`](@ref) ``$(_math(:G))``,
+namely the tangent space the tangent space ``T_{$(_math(:e))}$(_math(:G))`` at the [`Identity`](@ref),
```
mateuszbaran:
```suggestion
namely the tangent space ``T_{$(_math(:e))}$(_math(:G))`` at the [`Identity`](@ref),
```

src/Lie_algebra/Lie_algebra_interface.jl
```diff
@@ -0,0 +1,86 @@
...
+end
+
+_doc_lie_bracket = """
+    lie_bracket!(𝔤::LieAlgebra, X, Y)
+    lie_bracket!(𝔤::LieAlgebra, Z, X, Y)
+
+Compute the Lie bracket ``[⋅,⋅]: $(_math(:𝔤))×$(_math(:𝔤)) → $(_math(:𝔤))`` which fulfills
+
+1. ``[X,X] = 0`` for all ``X ∈ $(_math(:𝔤))``
+2. The Jacobi identity holds ``[X, [Y,Z]] = [[X,Y],Z] = [Y, [X,Z]]`` holds for all ``X, Y, Z ∈ $(_math(:𝔤))``.
```
mateuszbaran: "holds" appears twice here

src/group_actions/group_action_interface.jl
```diff
@@ -99,18 +99,19 @@ abstract type AbstractGroupAction{
...
 base_lie_group(::AbstractGroupAction)
 
 @doc """
     base_manifold(A::AbstractGroupAction)
 
-Return the $(_link(:AbstractManifold)) of the [`AbstractGroupAction`](@ref).
+Return the $(_link(:AbstractManifold)) the group action acts upon.
 """
-ManifoldsBase.base_manifold(::AbstractGroupAction)
+ManifoldsBase, base_manifold(::AbstractGroupAction)
```
mateuszbaran: ?

src/group_operations/addition.jl
```diff
@@ -0,0 +1,206 @@
+"""
+    AdditionGroupOperation <: AbstractGroupOperation
+
+A group operation that is realised introducing defaults that fall back
+to `+` and `-` being overloaded, for example
+`_compose(G::LieGroup{𝔽,AdditionGroupOperation}, a,b) = a+b`
```
mateuszbaran:
```suggestion
`_compose(G::LieGroup{𝔽,AdditionGroupOperation}, a, b) = a + b`
```

src/group_operations/addition.jl
```diff
@@ -0,0 +1,206 @@
...
+"""
+
+@doc "$(_doc_compose_add)"
+compose(::LieGroup{𝔽,AdditionGroupOperation}, g, h) where {𝔽}
+
+@doc "$(_doc_compose_add)"
+compose!(::LieGroup{𝔽,AdditionGroupOperation}, k, g, h) where {𝔽}
+
+function _compose!(G::LieGroup{𝔽,AdditionGroupOperation}, k, g, h) where {𝔽}
+    ManifoldsBase.copyto!(G, k, g + h)
```
mateuszbaran:
```suggestion
    k .+= g .+ h
```
This should be sufficiently generic to use here.

src/group_operations/addition.jl
```diff
@@ -0,0 +1,206 @@
...
+_doc_diff_left_compose_add = """
+    diff_left_compose(G::LieGroup{𝔽,AdditionGroupOperation}, g, h, X)
+    diff_left_compose!(G::LieGroup{𝔽,AdditionGroupOperation}, Y, g, h, X)
+
+Compute the differential of the left group multiplication ``λ_g(h) = g$(_math(:∘))h``,
+which simplifies for [`AdditionGroupOperation`](@ref) to ``Dλ_g(h)[X] = X``.
+"""
+
+@doc "$(_doc_diff_left_compose_add)"
+diff_left_compose(G::LieGroup{𝔽,AdditionGroupOperation}, g, h, X) where {𝔽}
```
mateuszbaran:
```suggestion
diff_left_compose(G::LieGroup{𝔽,AdditionGroupOperation}, g, h, X) where {𝔽} = X
```
IMO it makes sense to have it implemented like this to avoid allocations in the fallback.

src/group_operations/addition.jl
```diff
@@ -0,0 +1,206 @@
...
+Compute the differential of the inverse operation ``ι_{$(_math(:G))}(g) = g^-1 = -g``,
+which simplifies for [`AdditionGroupOperation`](@ref) to ``Dι_{$(_math(:G))}(g)[X] = -X``
+"""
+
+@doc "$(_doc_diff_inv_add)"
+diff_inv(G::LieGroup{𝔽,AdditionGroupOperation}, g, X) where {𝔽}
+
+@doc "$(_doc_diff_inv_add)"
+function diff_inv!(G::LieGroup{𝔽,AdditionGroupOperation}, Y, g, X) where {𝔽}
+    return copyto!(LieAlgebra(G), Y, -X)
```
mateuszbaran:
```suggestion
    Y .= (-1) .* X
    return Y
```
This should be sufficiently generic

src/group_operations/addition.jl
```diff
@@ -0,0 +1,206 @@
...
+_doc_diff_inv_add = """
+    diff_inv(G::LieGroup{𝔽,AdditionGroupOperation}, g, X)
+    diff_inv!(G::LieGroup{𝔽,AdditionGroupOperation}, Y, g, X)
+
+Compute the differential of the inverse operation ``ι_{$(_math(:G))}(g) = g^-1 = -g``,
+which simplifies for [`AdditionGroupOperation`](@ref) to ``Dι_{$(_math(:G))}(g)[X] = -X``
+"""
+
+@doc "$(_doc_diff_inv_add)"
+diff_inv(G::LieGroup{𝔽,AdditionGroupOperation}, g, X) where {𝔽}
```
mateuszbaran:
```suggestion
diff_inv(G::LieGroup{𝔽,AdditionGroupOperation}, g, X) where {𝔽} = -X
```

src/group_operations/addition.jl
```diff
@@ -0,0 +1,206 @@
...
+
+Compute the Lie group exponential on a [`LieGroup`](@ref) with an [`AdditionGroupOperation`](@ref).
+This can be computed in-place of `g`.
+
+Since `e` is just the zero-element with respect to the corresponding `+`, the formula reads ``g=0+X=X``.
+"""
+
+@doc "$(_doc_exp_add)"
+Base.exp(
+    ::LieGroup{𝔽,AdditionGroupOperation}, ::Identity{AdditionGroupOperation}, X, t
```
mateuszbaran:
```suggestion
    ::LieGroup{𝔽,AdditionGroupOperation}, ::Identity{AdditionGroupOperation}, X, t::Real
```

src/group_operations/addition.jl
```diff
@@ -0,0 +1,206 @@
...
+
+@doc "$(_doc_exp_add)"
+function ManifoldsBase.exp!(
+    G::LieGroup{𝔽,AdditionGroupOperation},
+    g,
+    ::Identity{AdditionGroupOperation},
+    X,
+    t::Number=1,
+) where {𝔽}
+    return copyto!(G, g, X)
```
mateuszbaran:
```suggestion
    g .= t .* X
    return g
```

src/group_operations/addition.jl
```diff
@@ -0,0 +1,206 @@
...
+Compute the Lie group exponential on a [`LieGroup`](@ref) with an [`AdditionGroupOperation`](@ref).
+This can be computed in-place of `g`.
+
+Since `e` is just the zero-element with respect to the corresponding `+`, the formula reads ``g=0+X=X``.
+"""
+
+@doc "$(_doc_exp_add)"
+Base.exp(
+    ::LieGroup{𝔽,AdditionGroupOperation}, ::Identity{AdditionGroupOperation}, X, t
+) where {𝔽}
```
mateuszbaran:
```suggestion
) where {𝔽} = t * X
```

src/group_operations/addition.jl
```diff
@@ -0,0 +1,206 @@
...
+Compute the inverse group element ``g^{-1}``, which for the [`AdditionGroupOperation`](@ref)
+simplifies to ``-g``. This can be done in-place of `h`.
+"""
+
+@doc "$(_doc_inv_add)"
+inv(G::LieGroup{𝔽,AdditionGroupOperation}, g) where {𝔽}
+
+@doc "$(_doc_inv_add)"
+function inv!(G::LieGroup{𝔽,AdditionGroupOperation}, h, g) where {𝔽}
+    return copyto!(G, h, -g)
```
mateuszbaran:
```suggestion
    h .= (-1) .* g
    return h
```

src/group_operations/addition.jl
```diff
@@ -0,0 +1,206 @@
...
+Compute the Lie group logarithm on a [`LieGroup`](@ref) with an [`AdditionGroupOperation`](@ref).
+This can be computed in-place of `X`.
+
+Since `e` is just the zero-element with respect to the corresponding `+`, the formula reads ``X=g-0=g``.
+"""
+
+@doc "$(_doc_log_add)"
+ManifoldsBase.log(
+    G::LieGroup{𝔽,AdditionGroupOperation}, ::Identity{AdditionGroupOperation}, q
+) where {𝔽}
```
mateuszbaran:
```suggestion
) where {𝔽} = q
```

src/interface.jl
```diff
@@ -160,116 +94,99 @@ end
...
     compose(G::LieGroup, g, h)
     compose!(G::LieGroup, k, g, h)
 
 Perform the group oepration ``g $(_math(:∘)) h`` for two ``g, h ∈ $(_math(:G))``
 on the [`LieGroup`](@ref) `G`. This can also be done in-place of `h`.
+
+!!! info
+    This function also handles the case where `g` or/and `h` are the [`Identity`](@ref)`(G)`.
+    Since this would lead to ambiguities when implementing a new group operations,
+    this function calls `_compose` and `compose!`, respectively, which is meant for the actual computation of
```
```suggestion
    this function calls `_compose` and `_compose!`, respectively, which is meant for the actual computation of
```
Probably?

src/interface.jl
```diff
@@ -437,46 +425,142 @@ function inv! end
...
+_doc_is_vector = """
+    is_vector(G::LieGroup, X; kwargs...)
+    is_vector(G::LieGroup{𝔽,O}, e::Indentity{O}, X; kwargs...)
+
+Check whether `X` is a tangent vector, that is an element of the |`LieAlgebra`](@ref)
+of `G`.
+The first variant calls [`is_point`](@extref ManifoldsBase.is_point) on the [`LieAlgebra`](@ref) `𝔤` of `G`.
+The second variant calls [`is_vector`](@extref ManifoldsBase.is_vector) on the $(_link(:AbstractManifold)) at the [`identity_element`](@ref).
+
+All kwyword arguments are passed on to the corresponding call
```
```suggestion
All keyword arguments are passed on to the corresponding call
```

src/interface.jl
```diff
@@ -437,46 +425,142 @@ function inv! end
...
+"""
+
+@doc "$(_doc_is_vector)"
+ManifoldsBase.is_vector(G::LieGroup, X; kwargs...) =
+    ManifoldsBase.is_point(LieAlgebra(G), X; kwargs...)
+
+@doc "$(_doc_is_vector)"
+function ManifoldsBase.is_vector(G::LieGroup{𝔽,O}, e::Identity{O}, X; kwargs...) where {𝔽,O}
```
```suggestion
function ManifoldsBase.is_vector(G::LieGroup{𝔽,O}, e::Identity{O}, X; kwargs...) where {𝔽,O<:AbstractGroupOperation}
```

src/interface.jl
```diff
@@ -437,46 +425,142 @@ function inv! end
...
+
+Check if points `g` and `h` from [`LieGroup`](@ref) are approximately equal.
+this function calls the corresponding $(_link(:isapprox)) on the $(_link(:AbstractManifold))
+after handling the cases where one or more
+of the points are the [`Identity`](@ref).
+All keyword argments are passed to this function as well.
+"""
+ManifoldsBase.isapprox(G::LieGroup, g, h; kwargs...) = isapprox(G.manifold, g, h; kwargs...)
+function ManifoldsBase.isapprox(
+    G::LieGroup{𝔽,O}, g::Identity{<:O}, h; kwargs...
```
```suggestion
    G::LieGroup{𝔽,O}, g::Identity{O}, h; kwargs...
```

src/interface.jl
```diff
@@ -437,46 +425,142 @@ function inv! end
...
+) where {𝔽,O<:AbstractGroupOperation}
+    return ManifoldsBase.isapprox(G.manifold, identity_element(G), h; kwargs...)
+end
+function ManifoldsBase.isapprox(
+    G::LieGroup{𝔽,O}, g, h::Identity{<:O}; kwargs...
```
```suggestion
    G::LieGroup{𝔽,O}, g, h::Identity{O}; kwargs...
```

src/interface.jl
```diff
@@ -437,46 +425,142 @@ function inv! end
...
+) where {𝔽,O<:AbstractGroupOperation}
+    return ManifoldsBase.isapprox(G.manifold, g, identity_element(G); kwargs...)
+end
+function ManifoldsBase.isapprox(
+    G::LieGroup{𝔽,O}, g::Identity{<:O}, h::Identity{<:O}; kwargs...
```
```suggestion
    G::LieGroup{𝔽,O}, g::Identity{<:O}, h::Identity{O}; kwargs...
```

src/interface.jl
```diff
@@ -437,46 +425,142 @@ function inv! end
...
+) where {𝔽,O<:AbstractGroupOperation}
+    return true
+end
+function ManifoldsBase.isapprox(
+    G::LieGroup{𝔽,O}, g::Identity{<:O}, h::Identity{<:O2}; kwargs...
```
```suggestion
    G::LieGroup{𝔽,O}, g::Identity{O}, h::Identity{O2}; kwargs...
```

src/interface.jl
```diff
@@ -530,5 +614,30 @@ function ManifoldsBase.log(G::LieGroup, e::Identity, g)
...
 end
 
+# explicit method error to avoid stack overflow
 @doc "$(_doc_log_id)"
-ManifoldsBase.log!(::LieGroup, X, ::Identity, g)
+function ManifoldsBase.log!(G::LieGroup, X, e::Identity, g)
+    throw(MethodError(ManifoldsBase.log!, (typeof(G), typeof(X), typeof(e), typeof(g))))
+end
+
+LinearAlgebra.norm(G::LieGroup, g, X) = LinearAlgebra.norm(G.manifold, g, X)
```
```suggestion
LinearAlgebra.norm(G::LieGroup, g, X) = norm(G.manifold, g, X)
```
We don't need to specify module for exported functions like `norm` when calling them.

src/interface.jl
```diff
@@ -530,5 +614,30 @@ function ManifoldsBase.log(G::LieGroup, e::Identity, g)
...
+
+function ManifoldsBase.representation_size(G::LieGroup)
+    return ManifoldsBase.representation_size(G.manifold)
```
```suggestion
    return representation_size(G.manifold)
```

src/interface.jl
```diff
@@ -530,5 +614,30 @@ function ManifoldsBase.log(G::LieGroup, e::Identity, g)
...
+end
+
+function Base.show(io::IO, G::LieGroup)
+    return print(io, "LieGroup($(G.manifold), $(G.op))")
+end
+
+function ManifoldsBase.zero_vector(
+    G::LieGroup{𝔽,O}, ::Identity{O}
+) where {𝔽,O<:AbstractGroupOperation}
+    return ManifoldsBase.zero_vector(G, identity_element(G))
```
```suggestion
    return zero_vector(G, identity_element(G))
```

src/interface.jl
```diff
@@ -530,5 +614,30 @@ function ManifoldsBase.log(G::LieGroup, e::Identity, g)
...
+end
+
+function ManifoldsBase.zero_vector!(
+    G::LieGroup{𝔽,O}, X, ::Identity{O}
+) where {𝔽,O<:AbstractGroupOperation}
+    return ManifoldsBase.zero_vector!(G, X, identity_element(G))
```
```suggestion
    return zero_vector!(G, X, identity_element(G))
```

test/LieGroupsTestSuite.jl/LieGroupsTestSuite.jl
```diff
@@ -0,0 +1,762 @@
+"""
+    LieGroupsTestSuite.jl
+
+This module provides tools and dummy structures to test functionality provided
+within `LieGroups.jl`.
+
+For every test function, several interactions to other functions can be activated.
+The following functions are expected to be available, since their defaults just pass through to the manifold
+* `is_point` both on the Lie group `G` and the Lie algebra `𝔤`
+* `isapprox(G,g,h)` and `issaprox(𝔤, X, Y)`
```
mateuszbaran:
```suggestion
* `isapprox(G, g, h)` and `isapprox(𝔤, X, Y)`
```

test/LieGroupsTestSuite.jl/LieGroupsTestSuite.jl
```diff
@@ -0,0 +1,762 @@
...
+"""
+    test_apply(A::AbstractGroupAction, g, p; expected=missing)
+
+Test  `apply`.
+
+# Keyword arguments
+* `expected=missing`: the result of the application of the group action.
+* `test_mutating=true`: test the mutating functions
+"""
+function test_apply(A::AbstractGroupAction, g, p; expected=missing, test_mutating=true)
```
mateuszbaran:
```suggestion
function test_apply(A::AbstractGroupAction, g, p; expected=missing, test_mutating::Bool=true)
```

test/LieGroupsTestSuite.jl/LieGroupsTestSuite.jl
```diff
@@ -0,0 +1,762 @@
...
+end
+
+"""
+    test_apply(A::AbstractGroupAction, g, p; expected=missing)
+
+Test  `apply`.
+
+# Keyword arguments
+* `expected=missing`: the result of the application of the group action.
+* `test_mutating=true`: test the mutating functions
```
mateuszbaran:
```suggestion
* `test_mutating::Bool=true`: test the mutating functions
```

test/LieGroupsTestSuite.jl/LieGroupsTestSuite.jl
```diff
@@ -0,0 +1,762 @@
...
+            @test isapprox(M, q1, q2)
+        end
+        !ismissing(expected) && @test isapprox(M, q1, expected)
+    end
+end
+#
+#
+# --- C
+"""
+    test_compose(G, g, h; kwargs...)
```
mateuszbaran:
```suggestion
    test_compose(G::LieGroup, g, h; kwargs...)
```

test/LieGroupsTestSuite.jl/LieGroupsTestSuite.jl
```diff
@@ -0,0 +1,762 @@
...
+
+Test  `compose` for given Lie group elements `g`, `h`.
+
+# Keyword arguments
+
+* `test_identity=true`: test that composing with the identity yields the identity (requires `identity_element`)
+* `test_inverse=true`: test that `g^{-1}g` is the identity (requires `inv`, `inv!`, and `is_identity`)
+* `test_mutating=true`: test the mutating functions
```
mateuszbaran:
```suggestion
* `test_identity::Bool=true`: test that composing with the identity yields the identity (requires `identity_element`)
* `test_inverse::Bool=true`: test that `g^{-1}g` is the identity (requires `inv`, `inv!`, and `is_identity`)
* `test_mutating::Bool=true`: test the mutating functions
```

test/LieGroupsTestSuite.jl/LieGroupsTestSuite.jl
```diff
@@ -0,0 +1,762 @@
...
+"""
+function test_compose(
+    G::LieGroup, g, h; test_inverse=true, test_identity=true, test_mutating=true
```
mateuszbaran:
```suggestion
    G::LieGroup, g, h; test_inverse::Bool=true, test_identity::Bool=true, test_mutating::Bool=true
```

test/LieGroupsTestSuite.jl/LieGroupsTestSuite.jl
```diff
@@ -0,0 +1,762 @@
...
+
+"""
+    test_conjugate(G::LieGroup, g, h; expected=missing)
+
+Test  `conjugate`.
+
+# Keyword arguments
+* `expected=missing`: the result of the conjugate can also be provided directly,
+  then neither `compose` nor `inv`  are not required.
+* `test_mutating=true`: test the mutating functions
```
mateuszbaran:
```suggestion
* `test_mutating::Bool=true`: test the mutating functions
```

test/LieGroupsTestSuite.jl/LieGroupsTestSuite.jl
```diff
@@ -0,0 +1,762 @@
...
+"""
+function test_conjugate(G::LieGroup, g, h; expected=missing, test_mutating=true)
```
mateuszbaran:
```suggestion
function test_conjugate(G::LieGroup, g, h; expected=missing, test_mutating::Bool=true)
```

test/LieGroupsTestSuite.jl/LieGroupsTestSuite.jl
```diff
@@ -0,0 +1,762 @@
...
+#
+# --- D
+"""
+    test_diff_apply(A::AbstractGroupAction, g, p, X; expected=missing)
+
+Test  `diff_apply`.
+
+# Keyword arguments
+* `expected=missing`: the result of the application of the group action.
+* `test_mutating=true`: test the mutating functions
```
mateuszbaran:
```suggestion
* `test_mutating::Bool=true`: test the mutating functions
```

test/LieGroupsTestSuite.jl/LieGroupsTestSuite.jl
```diff
@@ -0,0 +1,762 @@
...
+"""
+function test_diff_apply(
+    A::AbstractGroupAction, g, p, X; expected=missing, test_mutating=true
```
mateuszbaran:
```suggestion
    A::AbstractGroupAction, g, p, X; expected=missing, test_mutating::Bool=true
```

test/LieGroupsTestSuite.jl/LieGroupsTestSuite.jl
```diff
@@ -0,0 +1,762 @@
...
+end
+
+"""
+    test_diff_apply(A::AbstractGroupAction, g, p, X; expected=missing)
+
+Test  `diff_group_apply`.
+
+# Keyword arguments
+* `expected=missing`: the result of the application of the group action.
+* `test_mutating=true`: test the mutating functions
```
mateuszbaran:
```suggestion
* `test_mutating::Bool=true`: test the mutating functions
```

test/LieGroupsTestSuite.jl/LieGroupsTestSuite.jl
```diff
@@ -0,0 +1,762 @@
...
+"""
+function test_diff_group_apply(
+    A::AbstractGroupAction, g, p, X; expected=missing, test_mutating=true
```
mateuszbaran:
```suggestion
    A::AbstractGroupAction, g, p, X; expected=missing, test_mutating::Bool=true
```

test/LieGroupsTestSuite.jl/LieGroupsTestSuite.jl
```diff
@@ -0,0 +1,762 @@
...
+
+"""
+    test_diff_inv(G::LieGroup, g, X; expected=missing)
+
+Test  `diff_inv`.
+
+# Keyword arguments
+* `expected=missing`: the result of the differential of the inverse, if not provided,
+  only consistency between the allocating and the in-place variant is checked.
+* `test_mutating=true`: test the mutating functions
```
mateuszbaran:
```suggestion
* `test_mutating::Bool=true`: test the mutating functions
```

test/LieGroupsTestSuite.jl/LieGroupsTestSuite.jl
```diff
@@ -0,0 +1,762 @@
...
+"""
+function test_diff_inv(G::LieGroup, g, X; expected=missing, test_mutating=true)
```
mateuszbaran:
```suggestion
function test_diff_inv(G::LieGroup, g, X; expected=missing, test_mutating::Bool=true)
```

test/LieGroupsTestSuite.jl/LieGroupsTestSuite.jl
```diff
@@ -0,0 +1,762 @@
...
+        @test_throws DomainError copyto!(G, e, g)
+        return nothing
+    end
+end
+
+#
+#
+# --- E
+"""
+    test_exp_log(G, g, h, X)
```
mateuszbaran:
```suggestion
    test_exp_log(G::LieGroup, g, h, X)
```

test/LieGroupsTestSuite.jl/LieGroupsTestSuite.jl
```diff
@@ -0,0 +1,762 @@
...
+
+"""
+    test_copyto(G, g)
+
+Test that `copyto!` works also when copying over an `Identity`.
+
+The point `g` can be any point _but_ the `identity_element`.
+The group has to be a mutating one, that is, not work on isbit types.
+"""
+function test_copyto(G, g)
```
mateuszbaran:
```suggestion
function test_copyto(G::LieGroup, g)
```

test/LieGroupsTestSuite.jl/LieGroupsTestSuite.jl
```diff
@@ -0,0 +1,762 @@
...
+            @test isapprox(𝔤, Y1, Y2)
+        end
+        if !ismissing(expected)
+            @test isapprox(𝔤, Y1, expected)
+        end
+    end
+end
+
+"""
+    test_copyto(G, g)
```
mateuszbaran:
```suggestion
    test_copyto(G::LieGroup, g)
```

test/LieGroupsTestSuite.jl/LieGroupsTestSuite.jl
```diff
@@ -0,0 +1,762 @@
...
+        end
+    end
+    return nothing
+end
+
+#
+#
+# --- E
+"""
+    test_inv_compose(G, g, h, X)
```
mateuszbaran:
```suggestion
    test_inv_compose(G::LieGroup, g, h, X)
```

test/LieGroupsTestSuite.jl/LieGroupsTestSuite.jl
```diff
@@ -0,0 +1,762 @@
...
+                @test isapprox(𝔤, Y1, Y2)
+            end
+        end
+    end
+    return nothing
+end
+
+#
+#
+# --- E
```
mateuszbaran:
```suggestion
# --- I
```
Probably?

test/test_aqua.jl
```diff
@@ -0,0 +1,8 @@
+using Aqua, LieGroups, Test
+
+@testset "Aqua.jl" begin
+    Aqua.test_all(LieGroups; ambiguities=(broken=false, exclude=[
+        Base.:+, #ambiguities between Manifolds.Identity and LIeGroups.Identity
+        Base.:-, #ambiguities between Manifolds.Identity and LIeGroups.Identity
```
mateuszbaran:
```suggestion
        Base.:+, #temporary ambiguities between Manifolds.Identity and LieGroups.Identity
        Base.:-, #temporary ambiguities between Manifolds.Identity and LieGroups.Identity
```

test/test_interface.jl
```diff
@@ -0,0 +1,33 @@
...
+@testset "Generic Lie Group Interface functions" begin
+    M = LieGroupsTestSuite.DummyManifold()
+    op = LieGroupsTestSuite.DummyOperation()
+    G = LieGroup(M, op)
+    rs = "LieGroup(LieGroupsTestSuite.DummyManifold(), LieGroupsTestSuite.DummyOperation())"
+    @test repr(G) == rs
+    𝔤 = LieAlgebra(G)
+    op2 = LieGroupsTestSuite.DummySecondOperation()
+    rs2 = "Lie Algebra( LieGroup(LieGroupsTestSuite.DummyManifold(), LieGroupsTestSuite.DummyOperation()) )"
+    @test repr(𝔤) == rs2
```
mateuszbaran:
```suggestion
    rs2 = "LieAlgebra(LieGroup(LieGroupsTestSuite.DummyManifold(), LieGroupsTestSuite.DummyOperation()))"
    @test repr(𝔤) == rs2
```
I think that's how it should be printed.

src/interface.jl
```diff
@@ -437,46 +425,142 @@ function inv! end
...
+of `G`.
+The first variant calls [`is_point`](@extref ManifoldsBase.is_point) on the [`LieAlgebra`](@ref) `𝔤` of `G`.
+The second variant calls [`is_vector`](@extref ManifoldsBase.is_vector) on the $(_link(:AbstractManifold)) at the [`identity_element`](@ref).
+
+All kwyword arguments are passed on to the corresponding call
+"""
+
+@doc "$(_doc_is_vector)"
+ManifoldsBase.is_vector(G::LieGroup, X; kwargs...) =
+    ManifoldsBase.is_point(LieAlgebra(G), X; kwargs...)
```
```suggestion
    is_point(LieAlgebra(G), X; kwargs...)
```

src/interface.jl
```diff
@@ -437,46 +425,142 @@ function inv! end
...
+
+@doc "$(_doc_is_vector)"
+function ManifoldsBase.is_vector(G::LieGroup{𝔽,O}, e::Identity{O}, X; kwargs...) where {𝔽,O}
+    return ManifoldsBase.is_vector(G.manifold, identity_element(G), X; kwargs...)
```
```suggestion
    return is_vector(G.manifold, identity_element(G), X; kwargs...)
```

src/interface.jl
```diff
@@ -437,46 +425,142 @@ function inv! end
...
 end
 
-function diff_inv! end
-@doc "$_doc_diff_inv"
-diff_inv!(G::LieGroup, Y, g, X)
+_doc_inv_right_compose = """
+    inv_right_compose(G::LieGroup, h, g)
+    inv_right_compose!(G::LieGroup, k, h, g)
+
+Compute the inverse of the right group multiplication ``ρ_g(h) = h$(_math(:∘))g``,
```
```suggestion
Compute the inverse of the right group operation ``ρ_g(h) = h$(_math(:∘))g``,
```

src/interface.jl
```diff
@@ -437,46 +425,142 @@ function inv! end
...
+on the [`LieGroup`](@ref) `G`, that is Compute ``ρ_g^{-1}(h) = h$(_math(:∘))g^{-1}``.
```
```suggestion
on the [`LieGroup`](@ref) `G`, that is compute ``ρ_g^{-1}(h) = h$(_math(:∘))g^{-1}``.
```

src/interface.jl
```diff
@@ -437,46 +425,142 @@ function inv! end
...
-    diff_inv(G::LieGroup, g, X)
-    diff_inv!(G::LieGroup, Y, g, X)
+_doc_inv_left_compose = """
+    inv_left_compose(G::LieGroup, g, h)
+    inv_left_compose!(G::LieGroup, k, g, h)
 
-Compute the differential of the function ``ι_{$(_math(:G))}(p) = p^-1``, where
-``Dι_{$(_math(:G))}(p): $(_math(:𝔤)) → $(_math(:𝔤)).
-This can be done in-place of `Y`.
+Compute the inverse of the left group multiplication ``λ_g(h) = g$(_math(:∘))h``,
```
```suggestion
Compute the inverse of the left group operation ``λ_g(h) = g$(_math(:∘))h``,
```

src/interface.jl
```diff
@@ -374,47 +368,41 @@ See also [HilgertNeeb:2012; Definition 9.2.2](@cite).
...
-
-# See also
-
-[`identity_element`](@ref), [`identity_element!`](@ref)
-"""
-is_identity(G::AbstractDecoratorManifold, q)
+function ManifoldsBase.exp!(G::LieGroup, h, e::Identity, X, t::Number=1)
+    throw(
+        MethodError(
+            ManifoldsBase.exp!, (typeof(G), typeof(h), typeof(e), typeof(X), typeof(t))
```
```suggestion
            exp!, (typeof(G), typeof(h), typeof(e), typeof(X), typeof(t))
```

src/interface.jl
```diff
@@ -286,13 +203,35 @@ end
...
+function conjugate!(G::LieGroup, k, g, h)
     inv!(G, k, g) # g^{-1} in-place of k
-    compose!(G, k, h, h) # hk inplace of k
-    compose!(G, k, g, k) # gk inplace of k
+    compose!(G, k, h, k) # `h∘k` in-place of k
+    compose!(G, k, g, k) # `g∘k` in-place of k
     return k
 end
 
+ManifoldsBase.copyto!(G::LieGroup, h, g) = ManifoldsBase.copyto!(G.manifold, h, g)
```
```suggestion
ManifoldsBase.copyto!(G::LieGroup, h, g) = copyto!(G.manifold, h, g)
```

mateuszbaran (commented):
Next batch of comments :slightly_smiling_face:

docs/src/interface/operations.md
kellertuer: no, I meant that the operation together with the manifold builds a Lie group, In the sense of operation+Manifold=Lie group

src/Lie_algebra/Lie_algebra_interface.jl
kellertuer: Will remove the first.

src/group_actions/group_action_interface.jl
kellertuer:
```suggestion
ManifoldsBase.base_manifold(::AbstractGroupAction)
```

src/group_operations/addition.jl
kellertuer: My original idea was to (as often) document non-mutating, implement mutating, but sure we can also implement non-mutating here, just that this in one sense breaks the scheme: Since this does not allocate anew, it does return the X that comes in (instead of creating a copy).

src/group_operations/addition.jl
kellertuer: see above, most other allocating variants allocate, then this one does not, which might be confusing if you expect them to.

src/group_operations/addition.jl
kellertuer: I probably said this quite often now, I never know when to use copy and when to use `.=`, so good that you know.

src/interface.jl
kellertuer: For these I also never know when to put those and when not. The last 2 I added were when Aqua complained.

src/interface.jl
kellertuer: Ah, right, if we export them we do not, otherwise do. I sometimes di not fully have that in mind.

test/LieGroupsTestSuite.jl/LieGroupsTestSuite.jl
kellertuer: The alphabet. Also not so easy ;)

test/test_interface.jl
kellertuer: I Fixed this in a separate commit, since it also required to do the fix in the corresponding show-method.

src/interface.jl
kellertuer: I think this is still necessary to write, since otherwise the X does not seem to be filled correctly with zeros as far as I interpret most of the failing tests after these changes.

src/group_operations/addition.jl
kellertuer: I think this does not work as you think. For the two test `g1, g2 = [1.0, 0.0, 0.0], [0.0, 3.0, 0.0]` sure they sum has a zero third entry, but calling `compose` on them now yields

```
compose(G, g1, g2)
3-element Vector{Float64}:
 1.0
 3.0
 5.2936122611302e-310
```

Ah! The first plus is wrong!

src/interface.jl
kellertuer: Ah was not this maybe.

docs/src/notation.md
```diff
@@ -1,20 +1,18 @@
...
 | ``∘`` | a group operation | |
-| ``c:\mathcal G → \mathcal G, c_g`` | the conjugation map (with `g`) | |
+| ``c_g:\mathcal G → \mathcal G`` | the conjugation map (with `g`) | |
 | ``\mathrm{e}`` | identity element of a group | |
 | ``g, h, k`` | elements on a (Lie) group. Sometimes called points. | ``g_1, g_2, ...`` |
 | ``\mathfrak g`` | a Lie algebra | |
 | ``\mathcal{G}`` | a (Lie) group | |
-| ``λ_g: \mathcal G → \mathcal G`` | the left multiplication map ``λ_g(h) = g∘h`` | |
+| ``λ_g: \mathcal G → \mathcal G`` | the left group operation map ``λ_g(h) = g∘h`` | |
 | ``\mathcal M`` | a manifold | ``\mathcal M_1, \mathcal M_2,\ldots,\mathcal N`` | |
```
mateuszbaran: Maybe instead of mentioning things defined in Manifolds.jl notation file we should just link to it, and here only provide new definitions?

kellertuer: Right, we could do that as well for M and p and such.

kellertuer: Fixed it.

mateuszbaran (approved):

PR #12, opened 2024-10-24
Introduce the general linear group and the multiplication group operation.

### LINKED ISSUE #7: Transitioning from Manifolds.jl
@Affie wrote 
> > Ok, no breaking release, but let's face it: people will have to adapt from Manifolds.jl to LieGroups.jl. I call that breaking even if it is not.

> Yes please make it as easy as possible to upgrade, I understand some things will break, but even an error("deprecated, please use...) goes a long way. The v0.9 to v0.10 upgrade affects about 8 packages.

I was not aware that it is 8 packages. You can also stay on Manifolds.jl 0.10 for a while (0.11 in my mind for now just removes Lie groups then). Keep in mind that this package will also take a long long long while in my mind for now.
I am learning this along the way, try to document all functions with formulae, but I also have this thing called “semester”, lectures, supervision,... so I am not sure how long this package here takes until its first release – maybe Christmas? Easter? Something like that

The idea for transitioning to this package, for now I document all new functions (and where they come from) in the `NEWS.md` file (cf. https://github.com/JuliaManifolds/LieGroups.jl/blob/kellertuer/Intro/NEWS.md). I could start a branch of Manifold.jl and add a deprecation to every function I move, but that would probably double development time – and be a very tedious and boring coding task especially for all functions that really just move.

 I might write a tutorial for moving to cover the core parts that change and how to adapt. But feel free to propose what to do best to move.
And again, this is. probably not urgent, I would really not expect this to be any close to completion before Christmas – even starting this now was not my plan but maybe somewhen in a year originally, though I had the idea in a few issues prssibly about two years back, that “someone should do this”, just that there was no one.

kellertuer:
# 🛣️🗺️
* `GeneralLinearGroup`
  * [x] `check_point`, `check_vector`, `distance`, `embed`, `norm` (inherited from `InvertibleMatrices`)
  * [x] `get_coordinates`, `get_vector` in the `LieAlgebraOrthonormalBasis`
  * [x] `exp`, `exp!`, `log`,`log!` (will be inherited from matrix group operation
* [x] `AbstractMultiplicationGroupOperation`
  * [x] `MatrixMultiplicationGroupOperation`
  * [x] `exp`, `exp!`
  * [x] `log`, `log!`
  * [x] `*`, `/`, `\`
  * [x] `compose`
  * [x] `inv`
  * [x] `diff_inv` (formerly `inv_diff`)
  * [x] `diff_conjugate` 
  * [x] `diff_right_compose` and `diff_left_compose` (formerly `translate_diff`)
  * [x] `identity_element`, `identity_element!`
  * [x] `lie_bracket`, `lie_bracket!`  
* [x] Resolve #7 by using a table how to transfer function calls
* [x] document removal / replacement of `inverse_translate_diff`
* [x] `vee`
* [x] `hat`
* [x] `get_vector`
* [x] `get_coordinates`
* [x] `rand` and `rand!`
* [x] test coverage.


src/group_operations/multiplication_operation.jl
```diff
@@ -0,0 +1,295 @@
...
+```math
+Dι_{$(_math(:G))}(g)[X] = -g^{$(_tex(:transp))}Xp^{-1},
+```
+
+which stems from using the differential of the inverse from [Giles:2008](@cite) given by
+``D(g^-1)[X] = -g^{-1}Xg^{-1}`` composed with the push forward of the left composition
+``Dλ_$(_math(:e))(g)[X] = gX`` mapping from the Liea algebra into the tangent space at ``g``,
+and its adjoint ``D^*λ_$(_math(:e))(g)[X] = g^{$(_tex(:transp))}X``.
+Then we get ``g^{$(_tex(:transp))}(g^{-1}(gX)g^{-1})`` which simplifies to ``-g^{$(_tex(:transp))}Xp^{-1}`` from above.
+"""
```
kellertuer: @mateuszbaran could you maybe verify / check this? I think this should be correct. I only mention the Lie algebra case, because for now we assume all Lie groups represent their tangents in the Lie algebra (we can handle others when they appear). 
And I would even prefer having a reference for that derivation.

The one thing I do not like in this is, that the adjoint is only correct if the multiplicative Lie group we consider here has the Euclidean metric in the Lie algebra.

kellertuer: Maybe a generic `adjoint(𝔤::LieAlgebra, X)` could resolve the metric/adjoint problem.

kellertuer: I think I understood where its from, I just have to finde a good place to write this down.

If we have $Df(g) = L$ as a differential (linear map L) we have to change  _both_ “input” and “output” of that map, so if we write that out we can push forward any vector (from e to g before applying L) but we have to do the adjoint on the result (to pull back from g to e) afterwards.

We should try to write this down somewhere central maybe, or find a good reference.

kellertuer:
there is now only text coverage missing, so I already mark this as “Ready-to-Review”.

For testing the test suite has to be extended slightly, but besides that this PR is feature-complete.

src/group_operations/multiplication_operation.jl
mateuszbaran: I remember looking for a good reference for this but I couldn't find it.

> The one thing I do not like in this is, that the adjoint is only correct if the multiplicative Lie group we consider here has the Euclidean metric in the Lie algebra.

I think we can always have a specialized method for those non-Euclidean-metric cases.

> If we have D f ( g ) = L as a differential (linear map L) we have to change _both_ “input” and “output” of that map, so if we write that out we can push forward any vector (from e to g before applying L) but we have to do the adjoint on the result (to pull back from g to e) afterwards.

IIRC that's how it goes, yes.

kellertuer: Yes, we can surely leave this as is, then and specials / refine later.
One could do an `adjoint(...)` method for this case when refining this.

I also already documented that the tangent vectors for general linear are now in the Lie algebra and no linger in the embedding.

mateuszbaran:
Good work :+1: . One idea to discuss: should `GeneralLinearGroup` have `embed` and `project` for tangent vectors that left-multiply (or left-divide) by point? `GeneralUnitaryMatrices` in Manifolds.jl has that.

kellertuer:
Sounds reasonable. Do we want to stick to calling that “embed/project”? For me that is more of a change of a representer. We have the same `X` we just represent it once in the Lie algebra, once in the tangent space at `g`.

And in practice it is also just a call to the correct `diff_compose`, `diff_inv` combination, but sure, something nice to have which is easier to use.

So I am very much for introducing that here, but maybe we can have some better fitting names? Especially project does not fit so well for a change of a representer.

src/interface.jl
```diff
@@ -379,6 +390,95 @@ function ManifoldsBase.exp!(G::LieGroup, h, e::Identity, X, t::Number=1)
...
+
+@doc "$(_doc_get_vector)"
+ManifoldsBase.get_vector(G::LieGroup, g, c, B::ManifoldsBase.AbstractBasis)
+
+@doc "$(_doc_exp_id)"
+ManifoldsBase.get_vector!(G::LieGroup, X, g, c, B::ManifoldsBase.AbstractBasis)
+
+_doc_hat = """
+    hat(G::LieGroup, c)
+    hat(G::LieGroup, X, c)
```
```suggestion
    hat!(G::LieGroup, X, c)
```

src/interface.jl
```diff
@@ -627,6 +759,42 @@ function Base.show(io::IO, G::LieGroup)
     return print(io, "LieGroup($(G.manifold), $(G.op))")
 end
 
+_doc_vee = """
+    vee(G::LieGroup, X)
+    vee(G::LieGroup, c, X)
```
```suggestion
    vee!(G::LieGroup, c, X)
```

mateuszbaran:
> Sounds reasonable. Do we want to stick to calling that “embed/project”? For me that is more of a change of a representer. We have the same `X` we just represent it once in the Lie algebra, once in the tangent space at `g`.

`embed` and `project` were specifically meant to handle this conversion. Maybe it's not particularly intuitive but what would you propose instead? Note that we already have `change_representer` that is meant for dealing with cotangent vectors.

Affie:
Hi, I was looking at hat and vee and like the fact that it now works on the lie algebra without the extra parameter (point on manifold) that was in Manifolds.jl.
There's one thing I don't quite understand with vee and hat. Are they unique, ie. is there only one specific LieAlgebraOrthogonalBasis that can be used. Do you have a reference about vee and hat for met to look at?

mateuszbaran:
`vee` and `hat` are a simplification of `get_coordinates` and `get_vector` . LieGroups.jl assumes left-invariant tangent vector representation as a default so we can simplify it a bit further than in Manifolds.jl: the point is always the identity.

These two functions are definitely not uniquely defined, we just pick one reasonably common convention for each Lie group. If that convention doesn't work for some application, there is still `get_coordinates` and `get_vector` in their full generality. A formal definition of vee is for example in "Stochastic Models, Information Theory, and Lie Groups, Volume 2", section 10.2.4.

Affie:
Thanks, that helps. So, for us it's better to build against `get_coordinates` and `get_vector` with `LieAlgebraOrthogonalBasis` as the default basis to remain general but still get the "reasonably common convention" default.

Thanks for the reference. I didn't know about the diagram in 10.32. Is the diagram only relevant to vee, or is get_coordinates in general also dependent on the inner product in this way?

kellertuer:
> Hi, I was looking at hat and vee and like the fact that it now works on the lie algebra without the extra parameter (point on manifold) that was in Manifolds.jl.

Well, one idea of this package is to exactly do that: Rethink a few design choices and make them more “Lie-ish”. That takes a bit of time, but great to see, that this is recognised and valued.:)

> There's one thing I don't quite understand with vee and hat. Are they unique, ie. is there only one specific LieAlgebraOrthogonalBasis that can be used. Do you have a reference about vee and hat for met to look at?

As Mateusz said, similar to the `DefaultOrthogonalBasis` which is an opinionated choice – the LieOGB (or ONB? We can check) one is also just an opinionated choice of a basis. I would say that for example on SO(3) there is a very established generic choice – on others we just might stick to the one we feel is the most used. This choice, however, might be different from usual (Manifolds-jl-) choices, hence the extra name, to take that into account.

mateuszbaran:
> Thanks for the reference. I didn't know about the diagram in 10.32. Is the diagram only relevant to vee, or is get_coordinates in general also dependent on the inner product in this way?

Hm, I didn't pay much attention to that diagram before. It looks like a condition equivalent to orthonormality, and so far we only require orthogonality from vee and hat.

Readme.md
```diff
@@ -5,11 +5,15 @@
...
 </div>
 
-[![](https://img.shields.io/badge/docs-dev-blue.svg)](https://juliamanifolds.github.io/Manifolds.jl/latest/)
+[![](https://img.shields.io/badge/docs-dev-blue.svg)](https://juliamanifolds.github.io/LieGroups.jl/dev/)
 [![Code Style: Blue](https://img.shields.io/badge/code%20style-blue-4495d1.svg)](https://github.com/invenia/BlueStyle) [![CI](https://github.com/JuliaManifolds/LieGroups.jl/actions/workflows/ci.yml/badge.svg)](https://github.com/JuliaManifolds/LieGroups.jl/actions?query=workflow%3ACI+branch%3Amain) [![codecov.io](http://codecov.io/github/JuliaManifolds/LieGroups.jl/coverage.svg?branch=main)](https://codecov.io/gh/JuliaManifolds/LieGroups.jl/)
 
 This is a package to rework the Lie group features of [`Manifolds.jl`](https://juliamanifolds.github.io/Manifolds.jl/stable/) in a unified way into a separate package.
 
+> [!NOTE]
+> Since this is a rework of the features from [`Manifolds.jl`](https://juliamanifolds.github.io/Manifolds.jl/stable/), both `LieGroups.jl` and `Manifolds.jl` 0.10 export a few types. For that reason it's not recommended to load both packages in a productive environment, unless you are aware, which variables you want to import wherefrom.
```
mateuszbaran: I'd just say any production use of LieGroups.jl is not recommended until Manifolds.jl 0.11 is out.

kellertuer: I would not mind anyone using this earlier, just not together with Manifolds.jl (before 0.11).

mateuszbaran: Manifolds.jl is literally a requirement of LieGroups.jl so it's not really possible :slightly_smiling_face:

kellertuer: It is. As long as you just do

`using LieGroups`


all is fine, since LieGroups only exports its own `Identity` type for example but if you currently do

`using Manifolds, LieGroups` both export `identity` and you are doomed. That is what is meant here.

mateuszbaran: Then I'd suggest clarifying that you mean explicit import. Manifolds.jl is still loaded, the only difference is whether symbols are added to the module or not.

kellertuer: Well it already states that it is due to “exporting a few types” – maybe more precise would be “a few types of common name which leads to a naming conflict”.  But I can think about how to rephrase that sure.

kellertuer: I tried to rephrase it in the current commit.

mateuszbaran: Thanks, it is much better now :+1:

mateuszbaran:
Maybe it's also worth transferring `function log!(G::GeneralLinear{<:Any,𝔽}, X, p, q) where {𝔽}` and `function log!(::GeneralLinear{TypeParameter{Tuple{1}}}, X, p, q)` (and similarly for `exp`) from Manifolds.jl?

kellertuer:
I do not follow you here. Which ones are you referring to precisely? We have 3 variants of log nonetheless, but all 3 are covered

* the matrix logarithm is already given by the generic on in multiplication operation
* the Lie group log is given by fallback and the previous implementation (I hope, tests still pending)
* the Riemannian one will not explicitly be implemented here, but using `log!(base_manifold(G), X, p, q)` the original one can be accessed still.

similarly for `exp` the three cases hold.

mateuszbaran:
I mean for example this method: https://github.com/JuliaManifolds/Manifolds.jl/blob/cdfbe7f0a21129e55fffc070ba249faf96256cdc/src/groups/general_linear.jl#L206-L225 which I guess was implemented in such way for a reason? LieGroups.jl currently just computes the matrix logarithm.

mateuszbaran:
I've opened an issue to consider the issue later: #13. The one part to do here would be `𝔽 === ℝ && det(pinvq) ≤ 0 && throw(OutOfInjectivityRadiusError())` I think

kellertuer:
@Affie conquering your remark that lead to the note #7, what do you think about the table we now have in https://juliamanifolds.github.io/LieGroups.jl/previews/PR12/tutorials/transition/#Table-of-function-names-and-its-successors ?

docs/src/tutorials/transition.md
```diff
@@ -0,0 +1,52 @@
...
+| `apply_diff_group` | [`diff_group_apply`](@ref) | modifiers (diff/group) come first, consistent with [`ManifoldsDiff.jl`](https://juliamanifolds.github.io/ManifoldDiff.jl/stable/) |
+| | [`conjugate`](@ref), [`diff_conjugate`](@ref) | a new function to model ``c_g: \mathcal G → \mathcal G`` given by ``c_g(h) = g∘h∘g^{-1}`` |
+| `exp(G, g, X)` | `exp(`[`base_manifold`](@ref base_manifold(G::LieGroup))`(G), g, X)` | the previous defaults whenever not agreeing with the invariant one can now be accessed on the internal manifold |
+| `exp_inv(G, g, X)` | [`exp`](@ref exp(G::LieGroup, g, X, t::Number))`(G, g, X)`  | the exponential map invariant to the group operation is the default on Lie groups here |
+| `exp_lie(G, X)` | [`exp`](@ref exp(G::LieGroup, e::Identity, X, t::Number))`(G, `[`Identity`](@ref)`(G), X)` | the (matrix) exponential is now the one at the [`Identity`](@ref)`(G)`, since there it agrees with the invariant one |
+| `invervse_translate(G, g, h, c)` | [`inv_left_compose`](@ref)`(G, g, h)`, [`inv_right_compose`](@ref)`(G, g, h)` | compute ``g^{-1}∘h`` and ``g∘h^{-1}``, resp. |
```
Affie:
```suggestion
| `inverse_translate(G, g, h, c)` | [`inv_left_compose`](@ref)`(G, g, h)`, [`inv_right_compose`](@ref)`(G, g, h)` | compute ``g^{-1}∘h`` and ``g∘h^{-1}``, resp. |
```

Affie:
> @Affie conquering your remark that lead to the note #7, what do you think about the table we now have in https://juliamanifolds.github.io/LieGroups.jl/previews/PR12/tutorials/transition/#Table-of-function-names-and-its-successors ?

Thats perfect.

docs/src/tutorials/transition.md
```diff
@@ -0,0 +1,52 @@
...
+| `Manifolds.jl` | `LieGroups.jl` | Comment |
+|:---------- |:---------- |:-------------- |
+| `AdditionOperation` | [`AdditionGroupOperation`](@ref) | |
+| `LeftForwardAction` | [`LeftGroupOperation`](@ref)
+| `RightBackwardAction` | [`RightGroupOperation`](@ref) | |
+| `LeftBackwardAction` | [`InverseRightGroupOperation`](@ref) | note that this is now also aa [`AbstractLeftGroupActionType`](@ref) |
+| | [`LieAlgebra`](@ref)`(G)` | new alias to emphasize its manifold- and vector structure as well as for a few dispatch methods. |
+| `GroupManifold(M, op)` | [`LieGroup`](@ref)`(M, op)` | |
+| `RightForwardAction` | [`InverseLeftGroupOperation`](@ref) | note that this is an [`AbstractRightGroupActionType`](@ref) |
+| `adjoint` | [`adjoint`](@ref) | now implemented with a default, when you provide [`diff_conjugate!`](@ref).
```
Affie, line 31: Was this `adjoint_action`?

kellertuer: No, this was really just `adjoint`; this table does not yet cover much with actions, since actins have not yet been done (much) in this package, besides a first few thoughts on the interface.

kellertuer: See also https://juliamanifolds.github.io/LieGroups.jl/previews/PR12/interface/group/#Base.adjoint-Tuple{LieGroup,%20Any,%20Any} (but ignore the last sentence, I have to move that still).

Affie: > See also [https://juliamanifolds.github.io/LieGroups.jl/previews/PR12/interface/group/#Base.adjoint-Tuple{LieGroup,%20Any,%20Any}](https://juliamanifolds.github.io/LieGroups.jl/previews/PR12/interface/group/#Base.adjoint-Tuple%7BLieGroup,%20Any,%20Any%7D) (but ignore the last sentence, I have to move that still).

ok, I thought it was the adjoint_action because it matches https://juliamanifolds.github.io/Manifolds.jl/stable/manifolds/group.html#Manifolds.adjoint_action-Tuple{AbstractDecoratorManifold,%20Any,%20Any,%20Any}
From what I understand:
Wikipedia refers to this as the [adjoint representation or adjoint action](https://en.wikipedia.org/wiki/Adjoint_representation).
 "Stochastic Models, Information Theory, and Lie Groups, Volume 2" calls it the adjoint operator Ad(g).

kellertuer: Well, the `adjoint` here is a special case of `adjoint_action` if you want to be precise, namely when you consider the Lie groups group operation $\circ$ as a special case of a Lie group action, where the Lie group acts on itself (that is the manifold G acts on is M=G itself).
This is also the only group action that is currently available, since I want to have a few Lie groups and their operations finished before diving (back) into actions.

All of `adjoint`, `conjugate` and its `diff_conjugate` are new functions here, that I first and foremost needed for understanding, but that will also hopefully be useful later in defining a few new things. So maybe the answer to the initial question is maybe more like Nes or Yo ;)

kellertuer: My reference (hopefully used often) is Hilgert&Neeb https://link.springer.com/book/10.1007/978-0-387-84794-8 – mainly because a colleague of mine recommended that and I felt it was easy to read and reference to for me.

Affie: Ok, thank you.  I'll wait for the actions pr.  I'm tracking `adjoint_action` since I'm currently using it and figuring out a few related things.

kellertuer: Great! It might still take a while, since I would first like to make sure the groups and their group operations are safe and sound in their interface before continuing. but for that PR I have then in mind to ping you. The `adjoint_action` name will stay as before for the function name for sure.

kellertuer:
This should hopefully be finished by end of the week, the only few tests missing involve

* [x] `hat` and `vee`
* [x] new/simplified `get_vector` / `get_coordinates` on the LieAlgebra.

kellertuer:
Now this is really finished. 
With both addition and multiplication, the simpler Lie groups could now all be transferred.

src/interface.jl
```diff
@@ -559,6 +713,17 @@ function ManifoldsBase.isapprox(
...
+"""
+    lie_group_dimension(G::LieGroup)
+
+The dimension ``n=$(_tex(:rm, "dim"))_{$(_tex(:Cal,"G"))}`` of real space ``ℝ^n`` to which the neighborhood of
+each point of the [`LieGroup`](@ref) `M` is homeomorphic.
+
+This function is implemented to call [`manifold_dimension`](@extref `ManifoldsBase.manifold_dimension-Tuple{AbstractManifold}`) on the [`AbstractManifold`](@ref) `M`
+within the Lie group `G`.
+"""
+lie_group_dimension(G::LieGroup) = manifold_dimension(G.manifold)
```
mateuszbaran: Since Lie groups are still manifolds, I think this function isn't particularly useful? It would only cause confusion like "which one should I overload and use?".

src/interface.jl
```diff
@@ -627,6 +824,41 @@ function Base.show(io::IO, G::LieGroup)
     return print(io, "LieGroup($(G.manifold), $(G.op))")
 end
 
+_doc_vee = """
+    vee(G::LieGroup, X)
+    vee!(G::LieGroup, c, X)
+
+Compute the vee map ``(⋅)^∨`` that maps a tangent vector `X` from the [`LieAlgebra`](@ref)
+to its coordinates with respect to a certain bvasis in the Lie algebra
```
mateuszbaran: Maybe we should explicitly mention that we mean `LieAlgebraOrthogonalBasis`.

src/interface.jl
```diff
@@ -627,6 +824,41 @@ function Base.show(io::IO, G::LieGroup)
...
+
+```math
+X = $(_tex(:sum))_{i∈$(_tex(:Cal,"I"))} c_iB_i,
```
mateuszbaran, line 828: Maybe we could use {1, 2, ..., n} for simplicity instead of an index set? This function is only for finite-dimensional Lie groups after all.

mateuszbaran (commented):
A few comments

test/groups/test_general_linear_group.jl
```diff
@@ -0,0 +1,70 @@
...
+        @test exp(G, p, X) ≈ p * exp(X)' * exp(X - X')
+        q = exp(G, p, X)
+        Y = log(G, p, q)
+        @test Y ≈ X
+        @test exp(G, e, X) ≈ exp(X)
+        @test log(G, e, exp(X)) ≈ X
+        log(G, e, Identity(G)) == zeros(1, 1) # Matrix to matrix
+    end
+    # @testset "complex" begin
+    #     G = GeneralLinearGroup(1; field=ℂ)
```
mateuszbaran: This constructor doesn't work currently because `field` is a positional argument for `InvertibleMatrices` but `GeneralLinearGroup` only passes keyword arguments.

src/interface.jl
kellertuer: we could also just “pass down” then like 

```suggestion
manifold_dimension(G::LieGroup) = manifold_dimension(G.manifold)
```

? I would just also remove the doc string then.

src/interface.jl
kellertuer: Sure.

src/interface.jl
kellertuer: currently it is, so I wanted to be open for that, but sure we can also change that.

test/groups/test_general_linear_group.jl
kellertuer: oh, I missed that difference. I also missed to remove the commented out code.

kellertuer: Ah you added that code. Well, the only way to resolve this is to make field explicit instead of just having it in the kwargs and pass it down positional then.

src/interface.jl
kellertuer: For now I mentioned what the index set usually is. Will leave this open to be resolved by you if this is ok. That way we could easier move to the infinite dimensional case, which I might indeed explore at some point.

test/groups/test_general_linear_group.jl
kellertuer: I fixed it exactly that waz. Your tests are now added and work.

We could move them to also call the general test suite (mutable=false, only activate exp/log) though. What do you think?

src/interface.jl
mateuszbaran: OK. By the way, in the inifinite dimensional case all we can really do is picking some finite-dimensional subspace to represent, though that subspace is often adapted to the data we want to represent.

test/groups/test_general_linear_group.jl
mateuszbaran: Yes, that code was mine -- commented out to not report a failure on CI. I guess we can also change those tests to a call to `test_lie_group`.

src/interface.jl
mateuszbaran: Sure, that resolves this issue.

test/groups/test_general_linear_group.jl
kellertuer: Noticed we might have to reduce the number of points from 2 to 3 but that is also fine since until now we never use the third.

kellertuer: But we can also leave this as is for now, since it really just tests one function.

src/interface.jl
kellertuer: Numerically probably, mathematically not :)

mateuszbaran:
One more thing we might want to fix:
```julia
G = GeneralLinearGroup(3)
@test_throws ManifoldDomainError is_point(G, randn(ComplexF64, 3, 3); error=:error)
```
This used to work in Manifolds.jl but doesn't throw an error here.

kellertuer:
That might ne more something for Invertible matrices?

test/groups/test_general_linear_group.jl
mateuszbaran: I think it's sometimes easier to just test one or two functions instead of calling the whole testing procedure but I don't mind using a testing function instead. At least as long as it's not too time-consuming to call.

mateuszbaran:
I'm not really sure what goes wrong at the moment.

test/groups/test_general_linear_group.jl
```diff
@@ -0,0 +1,70 @@
...
+            diff_right_compose,
+            exp,
+            # hat, # requires a fix in Manifolds.jl to have an ONB on invertible matrices
+            inv,
+            inv_left_compose,
+            inv_right_compose,
+            is_identity,
+            lie_bracket,
+            log,
+            #rand, # requires a fix in Manifolds.jl to have rand on invertible matrices
```
kellertuer, line 33: If we do something on invertible matrices we can as well check to pass on bases to Euclidean and maybe check whether we find a way to implement rand.

test/groups/test_general_linear_group.jl
```diff
@@ -0,0 +1,72 @@
...
+            show,
+            #vee, # requires a fix in Manifolds.jl to have an ONB on invertible matrices
+        ],
+    )
+    expectations = Dict(
+        :repr => "GeneralLinearGroup(2; field=ℝ)", :lie_bracket => X1 * X2 - X2 * X1
+    )
+    test_lie_group(G, properties, expectations)
+
+    @test_broken is_point(G, Identity(G); error=:error)
```
mateuszbaran: One last thing that used to work in Manifolds.jl but doesn't here.

test/groups/test_general_linear_group.jl
mateuszbaran: Do you prefer to finish it in this PR or a new one?

docs/src/tutorials/transition.md
```diff
@@ -0,0 +1,52 @@
...
+| `log_inv(G, g, h)` | [`log`](@ref log(G::LieGroup, g, h))`(G, g, h)` | the logarithmic map invariant to the group operation is the default on Lie groups here |
+| `log_lie(G, g)` | [`log`](@ref log(G::LieGroup, e::Identity, g))`(G, `[`Identity`](@ref)`(G), g)` | the (matrix) logarithm is now the one at the identity, since there it agrees with the invariant one |
+| `switch_direction(A)` | [`inv`](@ref inv(::AbstractGroupAction))`(A)` | switches from an action to its inverse action (formerly the direction forward/backward, sometimes even left/right, do not confuse with the side left/right). |
+| `switch_side(A)` | [`switch`](@ref switch(::AbstractGroupAction))`(A)` | switches from a left action to its corresponding right action. |
+| `translate(G, g, h)` | [`compose`](@ref)`(G, g, h)` | unified to `compose` |
+| `translate_diff(G, g, X, c)` | [`diff_left_compose`](@ref)`(G, g, h, X)`, [`diff_right_compose`](@ref)`(G, g, h, X)` | for compose ``∘(g, h)`` the functions now specify whether the derivative is taken w.r.t. to the left (`g`) or right (`h`) argument |
```
mateuszbaran:
```
 ``∘(g, h)``
```
doesn't render correctly

test/groups/test_general_linear_group.jl
kellertuer: This is probably more something for Manifolds.jl anyways. so a PR there.

docs/src/tutorials/transition.md
kellertuer: Super interesting, it renders fine when I open the local one, the remote one on CI does not.

test/groups/test_general_linear_group.jl
kellertuer: Uff. I tried to fix this, because I do already provide the right `check_point` functions a bit, so I am not sure I have to maybe reimplement the whole dispatch path? Something along that is erroring :/

kellertuer: The usual thing when one tries to say: I am out of ideas: I saw the right idea: `check_size` was missing, but that is just nothing because identity is of right size anyways; check points were adapted for the positive (same op) and the negative case (different op).

kellertuer: Or maybe not...

Uff, this is complicated. It worked for short but now it even does not compile. I'll take a break.

docs/src/tutorials/transition.md
kellertuer: I have no clue what is happening here, I now had 3 variants that all render fine locally but not on CI.

kellertuer: Also variant 4 does not work, I do not know more variants to test, especially because all 4 render fine locally on my machine.

kellertuer: So. CI does seem to no longer update that page. I switched to ``g \circ h`` which is in math anyways the nice way to write it and it does not show upå. So this is a CI problem and will hopefully get better once we merge.

mateuszbaran (approved):
LGTM :+1:

docs/src/tutorials/transition.md
kellertuer: whatever it was, it renders fine on main https://juliamanifolds.github.io/LieGroups.jl/dev/tutorials/transition/ (and it renders on main at all after this PR :))

PR #14, opened 2024-11-04
Power, Product, and semidirect product manifolds 

kellertuer:
# ⛳️
Work out the three meta Lie groups: power, product, semi-direct product

> [!Note]
> The generalises are there, implementing all interface functions and tests for all three cases is still missing.

# 🛣️🗺️

* [x] implement the product Lie group – maybe as own struct?
* [x] check whether it is more reasonable for dispatch to indeed store a tuple of Lie groups in the Productmanifold. The outer operator should still be a tuple of operators or a properly typed product operation
* [x] implement the power Lie group, here the op is the actual (element wise) op, since we can dispatcher manifold. Maybe still as own structure?
* [x] also for power it might be better to store the Lie group inside for easier call to the subfiunctions – still the constructor should stay as the current ones are.
* [x] Implement a `default_|left|right]_action(G1, G2)` which is used in the semidirect product generation short notation.
* [x] implement functions for power manifold
* [x] implement functions for product manifold
* [x] find a general way to store a semi-direct product – maybe again as well based on two internally stored Lie groups.
* ~~implement functions for semidirect product manifold~~
* [ ] 📈test coverage


kellertuer:
I turned `runtests.jl` into a bash script so that you can run `test/runtest.jl` now from command line as well. It runs even `julia -O0` which on my machine reduces test time from `55 sec` to `21 sec`.

kellertuer:
Status: The main part of designing these three is done.

The remainder is to implement all functions form the interface for these.

kellertuer:
Oh, we might have an ambiguity problem in the current design:

```
  Candidates:
    identity_element!(PG::LieGroup{𝔽, Op, <:AbstractPowerManifold}, e) where {𝔽, Op}
      @ LieGroups ~/Repositories/Julia/LieGroups.jl/src/groups/power_group.jl:71
    identity_element!(::LieGroup{𝔽, AdditionGroupOperation, M} where M<:AbstractManifold{𝔽}, e) where 𝔽
      @ LieGroups ~/Repositories/Julia/LieGroups.jl/src/group_operations/addition_operation.jl:150
  
  Possible fix, define
    identity_element!(::LieGroup{𝔽, AdditionGroupOperation, M} where M<:(AbstractPowerManifold{𝔽, M} where M<:AbstractManifold{𝔽}), ::Any) where 𝔽
```

I have to think about how to best resolve this. For `_compose` it was easy: groups are dispatched on in `compose` (as well as `::Identity`), operators in `_compose`; maybe we need something similar for the identity element or the general architecture in general.

kellertuer:
I think for all three the scheme and modelling is done. The only thing left is to implement (and test) all interface functions for these three generic cases.

kellertuer:
@Affie since we discussed this in the last meeting: I now updated the transition page already.

Notably, the “new ways” to construct Lie groups – Given two Lie groups `G` and `H` are
* `G^n` for the power Lie group (or `PowerLieGroup(G, args; kwargs...)`, where all arguments and keywords are the same as for `PowerManifold`) – also no need to twiddle around with the operation.
* `G×H` for the `ProductLieGroup(G,H)` – again no need to twiddle around with operations or building a product manifold before
* `G⋉H` for the semidirect one, at least as long as the action has a reasonable default `default_left_action(G,H)` is defined, otherwise `LeftSemidirectProductLieGroup(G, H, a)` for your action `a` is the thing to use.

I hope this is more unified and nicer now :)

For this PR I am mainly missing implementing a few product and semidirect product function defaults.

kellertuer:
Tests are not yet fully running but all functions are in principle there for Power and Product Lie groups

the semidirect product is easier to finish on a next PR doing semidirect, SO(n) and SE(n), since then we have a concrete test case.

src/interface.jl
```diff
@@ -417,7 +409,9 @@ end
 
 # Fallback to a MethodError to avoid stack overflow
 @doc "$(_doc_exp_id)"
-function ManifoldsBase.exp!(G::LieGroup, h, e::Identity, X, t::Number=1)
+function ManifoldsBase.exp!(
+    G::LieGroup{𝔽,Op}, h, e::Identity{Op}, X, t::Number=1
+) where {𝔽,Op<:AbstractGroupOperation}
```
mateuszbaran, line 414: I think this was correct before: it's supposed to throw an error when group operations don't match.

kellertuer: Ah formerly it was just a fallback, but this here seems ring since Op should not be the same necessarily? Will check.

kellertuer: Hm, the only thing I can manage when I try to change this is to increase the number of ambiuities. If you see how to make this more general, feel free to do so.

mateuszbaran: Those ambiguities look real, that is `exp!` for `PowerLieGroup` likely won't work if the base point is `Identity`. I'm not sure if you'd like to make such test through the new framework or keep it separate?

kellertuer: Currently there are non if we leave the code here as is; I will check with power – and sure through the framework, that exp works with identity.

kellertuer: If I see that correctly the test with identity should already be running and working. cf https://github.com/JuliaManifolds/LieGroups.jl/blob/d16f091928f941f30daecbf419bc6551aec7334f/src/groups/power_group.jl#L204-L221, which does have coverage.

mateuszbaran: Hm, I see, though I'm not sure if this modified method actually can handle its original purpose. Anyway, maybe let's care about it when it actually happens.

kellertuer:
This PR should be finished now. There is one fallback on the interface not covered I think, but I do not know how to test it, maybe that fallback for inv and the identity is not necessary since for resolving an ambiguity it is implemented for every group separately anyways.

src/group_actions/group_action_interface.jl
```diff
@@ -85,41 +85,47 @@ $(_note_action_argument_order)
...
+* `type::T`: The type of the group action.
+* `group::L`: The group acting.
+* `manifold::M`: The manifold the group acts upon.
+
+See [HilgertNeeb:2012; Section 9.1.3](@cite) for more details.
 
-Return the $(_link(:AbstractManifold)) the group action acts upon.
 """
-ManifoldsBase.base_manifold(::AbstractGroupAction)
+struct GroupAction{T<:AbstractGroupActionType,L<:LieGroup,M<:ManifoldsBase.AbstractManifold}
```
mateuszbaran, line 117: Could you add somewhere the intended way to construct `GroupAction` for group operation action?  Should `manifold` be the Lie group itself or the underlying manifold?

mateuszbaran: I wanted to add a test for `switch(A::GroupAction) = GroupAction(switch(A.type), A.group)` and I wasn't sure what to do there.

EDIT: I've added a test based on dummy actions.

kellertuer: Nice, for such generic things I indeed wanted to use a few dummies.

kellertuer: For the `GroupOperationAction` That should be defined on `G` and not `M`, since we might have Lie groups on the same manifold with different group ops. and it should create a `GroupAction(GroupOperationActionType, G,G)`.

kellertuer:
For the last 6 missing lines: the 4 in semidirect will be easy with a concrete example, the 2 in interface I am not sure how to it.
But this is the coverage I was aiming for and the next PR would do the first concrete semidirect one.

src/group_actions/group_action_interface.jl
kellertuer: For the docs, it is documented at https://juliamanifolds.github.io/LieGroups.jl/previews/PR14/interface/actions/#LieGroups.GroupOperationAction-Tuple{AbstractGroupActionType,%20LieGroup}, but you are right, that could be improved, it also works or the other 3 cases (right, invleft, invright).

mateuszbaran: Thanks, I didn't think to check in that part of documentation.

mateuszbaran (approved):
LGTM

kellertuer:
Nice!

PR #17, opened 2024-12-13
(Special) Orthogonal / Unitary / Euclidean groups

### LINKED ISSUE #15: Keep point/vector representations In mind.
Similar to for example the `Hyperbolic` manifold, a Lie group might have different representations for its points and tangent vectors. This should be also the case here

Things to keep in mind

* `AbstractLieGroupPoint` might be a good idea for a subtype of the `ManifoldsBase.AbstractManifoldPoint` – mainly to emphasise a new kind of subtypes
* `AbstractLieAlgebraVector <: ManifoldsBase.TVector` analogue
* we should maybe define a default type that is equivalent to just using Arrays if resonable (or `PartitionArray` on ProductLieGroup`)
* maybe also something like an `GeneralLinearMatrix` or so in a general representation in matrix form?

kellertuer:
This PR aims to provide the special orthogonal group and the first concrete semidirect product: Special Euclidean, since the complex case is similar, this will also cover the special unitary case.

This PR resolves #15, since it introduces point and (Lie algebra) tangent vector types and carefully introduces all necessary passthroughs.

# 🛣️🗺️

* [x] implement `OrthogonalGroup`
* [x] Implement `SpecialOrthogonalGroup`
* [x] Implement `SpecialEuclideanGroup`
* [x] Implement `UnitrayGroup`
* [x] Implement `SpecialUnitaryGroup`
* [x] improve test coverage on the semidirect product groups.
* [x] read the documentation carefully since we changed `exp(G, Id, X)` to `exp(G,X)` and similarly `log(G, Ig, g)` to `log(G, g)`


kellertuer:
Status update 1: In theory SE(n) is defined, in practice I am not able to dispatch on the new const. Hence the constructor still falls back to calling the general show (of semidirect product) @mateuszbaran can you maybe take a look? After an hour of trying I am a bit out of ideas what is wrong with my constant `SpecialEuclideanGroup` definition.

mateuszbaran:
I've fixed the issue with SpecialEuclideanGroup alias :slightly_smiling_face:

kellertuer:
Nice. I would not have noticed the tuple necessary during the rest of the year.

You changed one . to a map. Should I do that for the rest as well?

kellertuer:
Oh and sorry for the capitalisation foo – Mac OS does not distinguish that on files when loading them … super annoying.

mateuszbaran:
> You changed one . to a map. Should I do that for the rest as well?

I generally prefer `map` over `.` when both are applicable because for the compiler it's much easier to optimize `map`.

> Oh and sorry for the capitalisation foo – Mac OS does not distinguish that on files when loading them … super annoying.

No problem, that was an easy fix :slightly_smiling_face:

kellertuer:
The main work today was a bit of structure, since now one of my main worries already works – `compose!` – the remaining functions are hopefully just stuff to transfer from manifolds.jl the todo list is basically all commented our functions in the two new test files.

So I am confident that this should not be too complicated – and am quite happy with how semidirect turned out.

src/groups/special_unitary_group.jl
```diff
@@ -0,0 +1,31 @@
...
+    )
+    return SpecialUnitaryGroup{typeof(GU).parameters[1]}(
+        GU, MatrixMultiplicationGroupOperation()
+    )
+end
+
+function Base.show(io::IO, G::SpecialUnitaryGroup)
+    size = Manifolds.get_parameter(G.manifold.size)[1]
+    return print(io, "SpecialUnitaryGroup($(size))")
+end
```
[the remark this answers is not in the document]
kellertuer: This is strange, seems like this can not be committed, locally my formatter also does not do this, I am lost what to do.

kellertuer:
All dummies are there, now the rest is mainly getting all functions from the test suite implemented and tested.

src/groups/orthogonal_group.jl
```diff
@@ -0,0 +1,283 @@
...
+
+#
+#
+# Generic special cases for O(n) and SO(n)
+
+_doc_exp_O2_id = """
+    exp(G::OrthogonalGroup{TypeParameter{Tuple{2}}}, ::Identity{MatrixMultiplicationGroupOperation}, X)
+    exp(G::SpecialOrthogonalGroup{TypeParameter{Tuple{2}}}, ::Identity{MatrixMultiplicationGroupOperation}, X)
+    exp!(G::OrthogonalGroup{TypeParameter{Tuple{2}}}, ::Identity{MatrixMultiplicationGroupOperation}, g, X)
+    exp!(G::SpecialOrthogonalGroup{TypeParameter{Tuple{2}}}, ::Identity{MatrixMultiplicationGroupOperation}, g, X)
```
kellertuer: @mateuszbaran 

I do agree, that dispatching only on the type parameter here is a great idea. It however makes the docs (that I am carefully revising currently) a bit clumsy. Do you maybe have an idea what to best do with these 4 lines (analogously for log – and for both also the cases of (S)O(3) and 4)?
My concern here is mainly readability of the docs, the code is fine as is. It should convey that it only works for the the parameter case but still be readable

src/groups/orthogonal_group.jl
```diff
@@ -0,0 +1,653 @@
...
+            E[1, 2] = β
+        end
+        copyto!(X, P * E * transpose(P))
+    else
+        det(q) < 0 && throw(
+            DomainError(
+                "The Lie group logarithm is not defined for $q with a negative determinant ($(det(q)) < 0). Point `q` is in a different connected component of the manifold $G",
+            ),
+        )
+        copyto!(X, real.(log(q)))
```
kellertuer: @Mateusz can you maybe take a Loog at the current error in one of the corner cases? I noticed I dod not take over `log_safe!`, maybe with the goal of simplifying the code. It seems to be the only difference of the 4x4 log compared to the Manifolds.jl implementation. I pent the whole morning trying to fix this but even _if_ I now copy over the (super long and techical) `log_safe!` cf https://github.com/JuliaManifolds/Manifolds.jl/blob/b52472bc116987e93b10f58b5e0599db4534728e/src/utils.jl#L83. Nothing works. I spent the whole morning trying to fix this. 
Sadly this code is more clever than I am, so I have no clue how to fix the one corner case failing.

src/groups/orthogonal_group.jl
mateuszbaran: Hm, it's somewhat problematic. I don't have any better idea for this docstring.

src/groups/orthogonal_group.jl
mateuszbaran: Sure, I will check it.

src/groups/orthogonal_group.jl
kellertuer: I will think a bit about it. Maybe we can leave out the types in the signature and only mention them in the text. In the current length, sure we can read it, but for everyone else it is a bit too technical. Spo maybe just `(G, e, g, X)` is fine enough if all of them are explained in the next few lines, also that G has a type parameter to dispatch on.

src/groups/orthogonal_group.jl
mateuszbaran: Fixed :slightly_smiling_face:

kellertuer: I would literally never have found that. Maybe I'll leave these technical special cases to others in the future then.
Thanks for fixing it. I actually removed a few get_vector get_coordinate calls in transferring the code, since it was sometimes cleaner afterwards.
I feel the same is the case here as well.

kellertuer: Hm, on CI exactly the same corner case (the third) is failing with the same error as yesterday.

mateuszbaran: Hm, weird, I think it worked for me locally but now when I try again, I get the same failure as on CI. Curiously, Manifolds.jl has the exact same issue when I modify the test to match your in LieGroups.jl. Numerical errors are enough to make it work :thinking:

kellertuer: I copied the tests over – especially the boundary cases, so I do not understand “when you modify the test to match”. I also do not understand the boundary cases not what you refer to with “numerical errors make it work”.

are the corner cases maybe wrong?

mateuszbaran: You switched `exp(X)` with `exp_lie(M, X)`. They are off by `eps()` in two places, which is enough to change the outcome but not enough to fail Manifolds.jl tests. I'm investigating.

kellertuer: Well, we do not _have_ `exp_lie` here with, so I can not do magic? I could do `exp(G, Identity(G), X)` which is its replacement, but then we compare A to A, which is equal? Well,...I was lost on this a few days ago already, that did not yet improve. But I also get too easily lost on those super-technical details. Will better not try to implement those again ;)

 I am looking forward to learning what the problem here then actually is.

mateuszbaran: I've looked at Gallier's paper about these computations. It looks like original implementation hoped that Julia's matrix `log` handles everything correctly except the case `α ≈ 0 && β ≈ π`. It clearly does not, and the failing test demonstrates the problem. I suggest marking this case as broken. I'm sure it could be fixed based on that paper but it looks like a work for a day or two.

kellertuer: Ok, that sounds fair. Actually just exp/log worked often fine before (when I tested the code here that is).
I think the case failing here is `α ≈ π/2 && β ≈ π` ? but sure, until someone (probably you) finds time for fixing that case, let's leave this one marked as broken.

src/groups/orthogonal_group.jl
kellertuer: The simplified version is now without any types and they are only mentioned in the text – that also reduces the signatures to 2 instead of 4 – one could for sure still make it 4 and mention the Lie groups again for completeness. Current for is something like (colouring from REPL not complete here after copy ;))

```julia
  exp(G, e, X)
  exp!(G, e, g, X)

  Compute the Lie group exponential function on the OrthogonalGroup \mathrm{O}(4) or SpecialOrthogonalGroup
  \mathrm{SO}(4), where e is the Identity{MatrixMultiplicationGroupOperation} and G uses a TypeParameter (@extref
  `ManifoldsBase.TypeParameter`) for dispatch.

  Similar to the 3×3 case, an efficient computation is provided, adapted from GallierXu:2002 (@cite),
  AndricaRohan:2013 (@cite) with a few numerical stabilisations.
```

src/groups/orthogonal_group.jl
mateuszbaran: Yes, that's exactly the failing case.

kellertuer:
To keep track of the remaining things for SE(n) mainly from the discussion and beginning rework from today.

* [x] introduce a `zero_vector(G, e, T)` providing a type `T` for the tangent vector (since `e` is the identity); same on the Lie algebra
* [x] introduce types for the representations
* [x] conversion between the representations
* [x] (maybe) write two new macros so we do not have to write “pass_through”s for the Lie group functions (@mateuszbaran can you maybe help set two similar macros up like the ones we have in ManifoldsBase? One for points one for vectors) – I can probably/hopefully fill in functions from the interface accordingly
* [x] `identity_element(G, T)` with type
* [x] `rand([rng], G, T)` (maybe also something for ManifoldsBase?); make the `rand(....; vector_at)`
* [x] `get_vector` and `hat` also still need the `T` trick
* [x] `exp`
* [x] `log`

mateuszbaran:
> * write two new macros so we do not have to write “pass_through”s for the Lie group functions (@mateuszbaran can you maybe help set two similar macros up like the ones we have in ManifoldsBase? One for points one for vectors) – I can probably/hopefully fill in functions from the interface accordingly

Do you mean something like `default_manifold_fallbacks`? That's probably the only one that needs a Lie group variant?

kellertuer:
That one definetly, and I am not so sure one needs additional / extended versions of `@manifold_element_forwards` and `@manifold_vector_forwards`?

I think I can fill one some parts, but setting it up I am not so sure, and you are maybe faster with that.

mateuszbaran:
OK, I will start something.

kellertuer:
Working through “maybe the type of tangent vector has to be provided” cases, I noticed that `get_vector` would maybe need to pass that down, maybe as a kwarg? then the internal `get_vector_lie` could “pick that up if it is provided”. But this would require a change in ManifoldsBase so I wanted to check for short.

The main problem is: Usually we can use `allocate_result` to find the type, even if we “only” have a point type ``::P`` we can infer the corresponding tangent point type (maybe the macro even introduces these allocations?). For `get_vector` here the point is the identity though, and that does not indicate the tangent vector type.

So when adding that to `ManifoldsBase` I could unify that and also do that for different point types in `rand`, which for tangent vectors has the same problem as above (as has `zero_vector`.

Let me know what you think, then I do `ManifoldsBase` next on this type issue.

kellertuer:
Small update:

Besides maybe `diff_`s all features should be done, besides `convert` between the two representations.

The main problem this PR now faces is that there is very very very very many ambiguities with identity,
since we have here a lot of functions that “do a default for identity” which leads to ambiguities with “implementations for specific point/vector types”.

In `ManifoldsBase.jl` we slit that by layer – the first on layer 1, the second, when necessary on Layer 2 on `_something(...)` methods. Would it make sense to introduce all the functions here with `_` as well? Might be a bit tricky to get right for a user who wants to implement a new Lie group, if they have to implement all the `_` functions.

Or we indeed do again an internal `exp_lie!` and `log_lie!` function to avoid ambiguities?
The other one is only `check_point`, since that is more like internal, we could indeed to a `_check_point` dispatch (where the inner one never has identity)

And the remaining ones are fighting all the cases of types that appear – currently often `submanifold_components` somewhere in `PowerManifold` to track down – but the ambiguities are my main remaining worry.

_edit:_ for check it might also be enough to pass to `check_point` on the Lie algebra. Will check the next days.

_edit2_: had a better idea, instead of checking always that identity fits the one in the Lie group (and run into ambiguities) we can leave it out - e.g. just `exp(G,X)` instead of `exp(G,e,X)`. Looks even nicer. Will check tomorrow how far one can push that idea. It should resolve most if not all ambiguities.

kellertuer:
Small update to the rambling from yesterday: The second edit seems a very good idea: It reduces code complexity, improves readability and make a lot of dispatches easer.

Far less code, redesign nearly done – a bit of check_point stuff left, but in the new code I am already below the number of failing tests as before. I also already updated the `transition.md` docs.

@mateuszbaran for `exp/exp!` on manifolds you always wanted `t::Number` to exist. Do you want that for the Lie group exponential `exp(G,X)` and `exp!(G, g, X)` (not one point less) as well? For now I did not yet introduce that there. but we probably could.

mateuszbaran:
> @mateuszbaran for `exp/exp!` on manifolds you always wanted `t::Number` to exist. Do you want that for the Lie group exponential `exp(G,X)` and `exp!(G, g, X)` (not one point less) as well?

Yes, as long as we consider Lie groups as `<:AbstractManifold`, they have to support the variants with `t`. The only choice consistent with existing API is to have a new function (`exp_lie`) for the methods that don't require the point. I think we've discussed this a few times already with simple examples like `exp(CircleGroup(), 2.0, 3.0)` being even logically ambiguous.

kellertuer:
Hm, as discussed in the chat. Then we need another solution here, still.

Will think about it again and then continue to work on it.

kellertuer:
This PR now requires https://github.com/JuliaManifolds/ManifoldsBase.jl/pull/221 to work / be merged.

This PR runs all its tests fine up to a presistency check due to the PR above missing,
and there is one final small thing left to fix.

in steps like this 
https://github.com/JuliaManifolds/LieGroups.jl/blob/5c853a2b569923b53731cba1fe2d3fc4a44f8a3f/src/groups/semidirect_product_group.jl#L215-L220

I would like the method to “write back” the result into the subcomponent of `k` but this does not seem to happen. This currently makes 4 remaining tests fail, but I am not able to figure out how to fix this. There is `_read` and `_write` for power manifolds/groups but it seems `submanifold_component` is equivalent to `_read` and there is no `_write` (with a `view`?)
@mateuszbaran can you maybe take a look, most of these super-nice technical tricks go back to your ideas.

Afterwards the tests should run fine and just checking for coverage and reading the docs super carefully are missing,
again since a lot of things changed here.

kellertuer:
I think I understand the problem of the last 3 tests failing:

the `submanifold_component` I implemented for `AbstractArray` and `:Rotation` for example returns a view, which usually these functions do not. The code above accesses `1` and `2` (not `:Rotation` and `:Translation`) hence does not return views and hence can not write back.

Should we maybe experiment first here in `LieGroups` with a `submanifold_component_vew`? In the long run that could maybe replace `_write` in `ManifoldsBase.jl`.

_edit:_ For the commit below this comment I did a proof-of-concept “misusing” the `submanifold_component` to return views.
Previously on `AbstractMatrix` it would fail, now it works fine, but I seem to have missed a view on the extension (`ArrayPartition`case) by now.
Let's first decide what best to do before I try to fix both.

mateuszbaran:
I think your `submanifold_component` methods are fine. They provide access to valid representations of components. I don't think we need a `_read`/`_write` distinction for `submanifold_component`, all access can be of the `_write` type.

kellertuer:
Then we should maybe aim to get that consistent to a `_write`-like style?

Besides that I then just have to figure out where the `ArrayPartition` cases fail ;)

mateuszbaran:
What's actually inconsistent? Which methods behave differently?

kellertuer:
If I just “pass on” to `G.manifold` I think they are not views and will not write back, but I can check. Also I thought, the reason that `_write` exists was because `subamifold_component` was not meant to be “writable access”?

Feel free to experiment with the definitions in the extension yourself – some of them seem to not yet “write back”, so they might have to still return views I think.

mateuszbaran:
If you pass on to `G.manifold`, those matrices are no longer valid points AFAIK.

> Also I thought, the reason that `_write` exists was because `subamifold_component` was not meant to be “writable access”?

`_write` is for `PowerManifold` to deal with the case of arrays of immutable elements. We support a similar thing differently in `ProductManifold` (by implementing non-in-place method).

kellertuer:
I see, then until now it did not matter whether `submanifold_component` was “writable” or not. I think it is at least not consistently writeable. I am also not sure we can get it consistently so.
However, that is the last point that needs to be fixed here for the tests :)

mateuszbaran:
The failing tests are for the `ArrayPartition` representation; I'm not sure how it could be related to `submanifold_components` not working right if it just gives the component arrays which are writable. And the failing tests aren't even for the mutating methods.

kellertuer:
Sure, the two methods called (exp/log without a point) they allocate and call the mutating ones. For me it looks like they are because the numbers are the same as for the case where AbstractMatrix did not “write back” the application of the action. (for exp/log which internally use 2 x compose and 1 x apply).
But I have a bit time on Wednesday to dive deeper into the code probably.

The tests afterwards that directly call the mutating functions just check that these give the same results as the allocating ones, to they produce the same (wrong) results currently)

mateuszbaran:
If the output wasn't mutable, you'd get a runtime error or uninitialized memory, not a somewhat reasonably looking answer like I see. I'm 99% sure it's an unrelated logic error.

kellertuer:
So for AbstractMatrix the problem was that there was a `return A[1:n,1:n]` that allocated a new matrix and returned that and the `apply`hence happened in place of that new matrix and not the original one which was then used in the following compose. 

That does not result in an error just that the in-place computation is done “in the wrong place” and hence never written back. Maybe something like that is also currently happen, at least the numbers are the same wrong ones as in the one that I fixed before.

mateuszbaran:
`submanifold_component` is already writable for `ArrayPartition`:
```julia
julia> using Manifolds, RecursiveArrayTools

julia> M = ProductManifold(Sphere(2), Sphere(3))
ProductManifold with 2 submanifolds:
 Sphere(2, ℝ)
 Sphere(3, ℝ)

julia> p = rand(M)
([-0.893222710741945, -0.008594280645648567, -0.4495323429465287], [0.6752061658308836, 0.005200949448816773, -0.2518961443382168, 0.6932661222188261])

julia> scp_a, scp_b = submanifold_components(M, p)
([-0.893222710741945, -0.008594280645648567, -0.4495323429465287], [0.6752061658308836, 0.005200949448816773, -0.2518961443382168, 0.6932661222188261])

julia> scp_a .= 0
3-element Vector{Float64}:
 0.0
 0.0
 0.0

julia> scp_b .= 0
4-element Vector{Float64}:
 0.0
 0.0
 0.0
 0.0

julia> p
([0.0, 0.0, 0.0], [0.0, 0.0, 0.0, 0.0])
```

kellertuer:
Then I have zero idea where even to start to pin down that error. For the abstract matrix case it took me 3-4 hours to dive into the code, so I will do that then Wednesday probably again for ArrayPartition. From outside it looks like the same error because it is the same numbers as if apply was not applied. If it is not – then that is for some other reason that is looks like that.

But I can not dive into that today.

kellertuer:
Step (1) I found a problem in `compose` for semidirect groups, where `compose!(G, h, g, h)` would “overwrite” one element im h top early (same for `compose!(G, g, g, h)`).

Now Left on Partition works
For Right, there seems to still be a sign error in the matrix Lie group exponential (hence also the CS one is wrong still). I could bring that down to `exp(G, g, -X)` would yield the same as the matrix multiplication – oh not completely, there is a mother sign error then in the T(n) component.
Not sure why, we only have one implementation for both currently (left&right); Maybe I confused exactly some left/right for them. If they need more than one unified implementation this also holds for the cases 3,4,n.
Is maybe now more a math problem than the one I solved now.

kellertuer:
Found it. I was not clever enough to provide a correct Lie algebra tangent vector 🙄

But now finally tests run fine, so only reading the docs, and improving test coverage (once ManifoldsBase.jl 1.0 is available and Manifolds.jl has been adapted to that) are left to do.

This PR is _finally_ feature complete.

mateuszbaran:
Nice, it's great that you've managed to figure it out :+1:

kellertuer:
This PR only misses the code coverage check. For that we need both
* [x] the merge of https://github.com/JuliaManifolds/ManifoldsBase.jl/pull/221 and the new version tagged  (which seems nearly finished)
* [x] the corresponding merge https://github.com/JuliaManifolds/Manifolds.jl/pull/781 to be merged and a new version registered (locally its tests already run fine)

kellertuer:
Nice! CI is running again, so the only thing left on this PR is to fix the remaining uncovered lines of tests.

kellertuer:
Uff, the types points and passing them on transparently is quiiiite a headache. Spent an hour to fix one of many errors left for just the first of the two sets of point types we have (and further fallbacks that need to be introduced)... This will need more time than I expected

kellertuer:
Hm, I have an ambiguity problem again, can you maybe check whether you have an idea @mateuszbaran ?

The problem is the following: For `exp(G, g)` on special Euclidean we have
* a point type and its fallback – hence a typed `g` for those
* specialisations on the size f the Lie group, hence typed variants for `G`.

For them
* the first already has two variants one for the left semidirect one for the right semidirect variant
* the second has 3 variants: size `n=2,3` and the generic one, but these work (thanks to submanifold_component) for any representation basically.

But this also yields 2 times 3, so 6 ambiguous definitons.

How to best resolve this?
a) Do left/right _and_ n=2,3,any fallback marcos? That would at least only be 6 macro calls, though the Lie group types get super involved and longish
b) implement the efficient variants for ArrayPartition and abstract matrix separately (so double the current code but omit submanifold_component basically)?
c) [...some other magic? ...]

I am tending a bit towards b) without any good reason.

src/interface.jl
```diff
@@ -904,9 +904,26 @@ macro default_lie_group_fallbacks(TG, TP, TV, gfield::Symbol, Xfield::Symbol)
             return k
         end
 
+        function LieGroups.exp(G::$TG, X::$TV)
```
mateuszbaran: This shouldn't be needed in this macro (the ManifoldsBase one should cover exp and log).

kellertuer: ManifoldsBase never covers the case with one argument, that is a new variant here in LieGroups.

kellertuer: eh 2 arguments, so the one without a point I mean.

mateuszbaran: OK, so it's this... are you sure it's less problematic than requiring `::Identity`?

kellertuer:
Luckily after exp/log I hope it is only hat/vee that are left to fix (though the tests for the other type wrapper are not yet present).

src/interface.jl
kellertuer: So with Identity I did not run in the 6 ambiguities above but about 200 others :D

kellertuer: Or phrased differently, with `:Identity` we basically need a split like for `compose` so a `_compose` / `_exp`. Then this is really just easier, since without the `t` now this is always completely clear what is meant for `exp(M, p, X)` vs `exp(G, X)`, so no semantic ambiguity – for the types see above, I think those would even appear in ManifoldsBase if someone has a type wrapper and specialises on size

mateuszbaran:
> How to best resolve this? a) Do left/right _and_ n=2,3,any fallback marcos? That would at least only be 6 macro calls, though the Lie group types get super involved and longish b) implement the efficient variants for ArrayPartition and abstract matrix separately (so double the current code but omit submanifold_component basically)? c) [...some other magic? ...]

Variant (b) sounds good. The only magic I could suggest is implementing the generic logic in a helper function that is called by the separate methods for `ArrayPartition` and `AbstractMatrix`, which is essentially (b) anyway.

kellertuer:
Nice, then I will go for b) the change is also not too large, just extracting the “subparts” will be different and the other lines a bit of duplication (a bit apart since the ArrayPartition one is in an extension).

kellertuer:
...and maybe more general a design tipp should be: If you specialise on the manifold size, maybe think about also specialising on the point/vector types and make it fully declared (to avoid the current point I am at).

But sure, then I know what to do – maybe later today, maybe tomorrow.

kellertuer:
I fixed most cases I think, just two ambiguities left, that I have to think about a bit

* `zero_vector(𝔤, ::SpecialPointType)` is not so easy to define in the macro since we do not have the type of the LieAlgebra. Maybe one can trick that, since we have the type of `G`.
* similarly there are a few cases like `log!(G, X::SpecialTangentVectorType, e::Identity)` where I currently still get an ambiguity when trying to pass through, since `e` is “just” the idenity and not `Identity{O}` with the Operation being the one from G – again in the macro we only have the type of G, not its group op type.

For both I ran out of ideas for today. For the rest, the idea “if you specialise on the dimension, also specialise point/vectors” works fine and even removes the necessity to introduce a few log/Identity special dispatches, I think.

mateuszbaran:
I think I've managed to fix those issues by introducing additional parameters for the macro. It now knows the Lie algebra. I've also fixed the failing Aqua test

kellertuer:
Hm, i had it down to 3 Tests faulig now it is more Tests again failing; but sure Addict those is an idea; LieAlgebra seems a bit generic in the Marco call maybe.

But I can continue from there Tomorrow afternoon probably.

mateuszbaran:
Sometimes a failing test may disable other failing tests (which are then not reported) so it's better to measure progress by the number of passing tests.

kellertuer:
Yeah, you're right – sometimes after a long time of fixing bugs/tests I forget about that :)

Also with a fresh look, since you “carry down” Group, Field and Op, the Lie algebra thing is indeed nice! The macro is now maybe a bit crowded to call, but well.

kellertuer:
Hm the Lie algebra pass throughs are really tricky. I just spent 2 hours – just to try to get `isapprox(𝔤, X, Y)` maybe to get to “unwrap” its X and Y – I am not able to define that so that it works and I am out of ideas why my definitions are never hit. It seems independent of whether I define them or not – they are never hit (dispatched on). State see last commit.

I try to work on this every day, but for this week the progress is really like last week – zero.

kellertuer:
Uff, that was again challenging, but I hope the “passthrough” now works – what is missing is all the special cases: Type-size dispatch, ArrayPartition extension,...

I did split the macro into two to reduce parameter length
* Lie group now only needs the op
*  Lie algebra does not need points.

that way both macros are a bit shorter to call. I also unified them with the manifolds one in the sense that instead of defining `exp` they now define the necessary allocation.

kellertuer:
By now this PR will also resolve #15

kellertuer:
I did all tests on SE(n) now, just the generic implementation of inv is missing on semidirect product.

Maybe I‘ll check a few of the other missing tests as well afterwards. But with most of the testing finished (90% of patch) I‘ll already mark this as ready-to-review.

mateuszbaran:
OK, nice, I will check the PR soon-ish.

docs/src/tutorials/transition.md
```diff
@@ -23,38 +23,38 @@ The list is alphabetical, but first lists types, then functions
...
 | `apply_diff_group` | [`diff_group_apply`](@ref) | modifiers (diff/group) come first, consistent with [`ManifoldsDiff.jl`](https://juliamanifolds.github.io/ManifoldDiff.jl/stable/) |
 | | [`conjugate`](@ref), [`diff_conjugate`](@ref) | a new function to model ``c_g: \mathcal G → \mathcal G`` given by ``c_g(h) = g∘h∘g^{-1}`` |
-| `exp(G, g, X)` | `exp(`[`base_manifold`](@ref base_manifold(G::LieGroup))`(G), g, X)` | the previous defaults whenever not agreeing with the invariant one can now be accessed on the internal manifold |
-| `exp_inv(G, g, X)` | [`exp`](@ref exp(G::LieGroup, g, X, t::Number))`(G, g, X)`  | the exponential map invariant to the group operation is the default on Lie groups here |
-| `exp_lie(G, X)` | [`exp`](@ref exp(G::LieGroup, e::Identity, X, t::Number))`(G, `[`Identity`](@ref)`(G), X)` | the (matrix) exponential is now the one at the [`Identity`](@ref)`(G)`, since there it agrees with the invariant one |
+| `differential_exp_argument_lie_approx` | - | is discontinued, see the `differential_exp_argument` in `ManifoldDiff.jl` instead, since `exp_lie` is now just `exp`. |
```
mateuszbaran: It's currently present in `ManifoldDiff.jl`, why would we discontinue it?

src/Lie_algebra/Lie_algebra_interface.jl
```diff
@@ -42,23 +42,228 @@ function LieAlgebra(G::LieGroup{𝔽,O}) where {𝔽,O<:AbstractGroupOperation}
...
+        c,
+        identity_element(G, T),
+        X,
+        ManifoldsBase.DefaultOrthogonalBasis(N),
+    )
 end
-function ManifoldsBase.get_vector!(𝔤::LieAlgebra, X, c, B::ManifoldsBase.AbstractBasis)
+
+_doc_get_vector = """
+    get_vector(G::LieGroup, c, B::AbstractBasis; kwargs...)
```
mateuszbaran, line 126: What `kwargs` does this support?

mateuszbaran: Ah, I see, `tangent_vector_type`.

src/Lie_algebra/Lie_algebra_interface.jl
```diff
@@ -42,23 +42,228 @@ function LieAlgebra(G::LieGroup{𝔽,O}) where {𝔽,O<:AbstractGroupOperation}
...
+"""
+
+# function hat end
+@doc "$(_doc_hat)"
+function hat(𝔤::LieAlgebra{𝔽}, c) where {𝔽}
+    return get_vector(𝔤, c, DefaultLieAlgebraOrthogonalBasis(𝔽))
+end
+function hat(𝔤::LieAlgebra{𝔽}, c, T::Type) where {𝔽}
+    return get_vector(𝔤, c, DefaultLieAlgebraOrthogonalBasis(𝔽); tangent_vector_type=T)
+end
```
mateuszbaran: Why do you try using non-real-coefficient bases here? I'd prefer real-coefficient basis to be the default.

docs/src/tutorials/transition.md
kellertuer: We can check whether we have a good way to have the `_approx` again?

The `_lie_` part is no longer necessary/useful since `exp_lie` does not longer exist and is called `exp`,
so sure we can rephrase this here that (up to the approx) the is now called `differential_exp_argument`.

src/Lie_algebra/Lie_algebra_interface.jl
kellertuer: So basically replace F with R? Fine for me, until now I was just going for the more generic case, since that was what was done before.

docs/src/tutorials/transition.md
mateuszbaran: The `lie` in the name refers to the fact that this approximation is only for Lie groups. We can rename it if you prefer but the functionality seems to me important enough to have in some function.

src/Lie_algebra/Lie_algebra_interface.jl
kellertuer: We could try to do it implicitly as well. as we do in `Manifolds.jl` in a few places, but the main problem compared to `Manifolds.jl` is that for manifolds we have some point `p` where we are at and we can use its type to “find” (or infer) the corresponding tangent vector type. Herer we do not have that and I do expect `c` to really always just be an array of coefficients.

src/Lie_algebra/Lie_algebra_interface.jl
mateuszbaran: Yes. I think the cases where someone might want complex coefficients are rare enough that they should not be the default.

docs/src/tutorials/transition.md
kellertuer: But that should come from the fact that the first argument of such a function is a Lie group? So why should that then be in the function name (repeated)?

src/Lie_algebra/Lie_algebra_interface.jl
kellertuer: But that is not the default here, it would just take the coeffs of the Lie group. 
However, we can surely change the default to always be real coefficients if you prefer; then the basis would be(however) be larger in dimension (the one of the complex Lie group)?

For now this is really just the old default I think, but sure we can change that to the more common case.

docs/src/tutorials/transition.md
mateuszbaran: We don't have a dispatch on the approximation method so the easiest way to introduce new functionality is by having a new function name. `differential_exp_argument` is currently not meant for approximations. I'm open to renaming but figuring out how to use the name `differential_exp_argument` for this is a lot of useless work IMO.

src/Lie_algebra/Lie_algebra_interface.jl
mateuszbaran: We should never have the coeffs of the manifold dictate the coefficients of the (default) basis of tangent spaces. The basis would be larger but the group having complex coefficients doesn't mean it's a complex manifold (which is a fairly strict requirement; for example the complex circle is not a complex manifold and does not have a complex coefficient basis).

docs/src/tutorials/transition.md
kellertuer: We could go for `differential_exp_argument_approx`? I am not sure whether to put that here or in ManifoldDiff (maybe just here first then), my main argument would be to not use `_lie` in the name

The other argument this is currently discontinued is that I was also too lazy what it actually did back then – but I would be fine introducing it as an approx variant without the lie in the name, sure.

src/Lie_algebra/Lie_algebra_interface.jl
kellertuer: Sure, then let's change the default here.

docs/src/tutorials/transition.md
mateuszbaran: Maybe let's leave it for later. It's currently in ManifoldDiff and we can discuss any changes separately from this PR.

mateuszbaran: For this PR I'd suggest just sticking to facts: it's not discontinued right now and it's available in ManifoldDiff.

kellertuer: Well it is discontinued to the extend that it was defined for `GroupManifolds` and is not defined for `LieGroups` – and we should probably rename it in ManifoldDiff at some point then. 


But we can rephrase this entry in the list of changes to make that clearer, that is correct.

mateuszbaran: Yes, right, it needs to be updated to work with LieGroups.jl and it's useful to note that.

src/group_operations/addition_operation.jl
```diff
@@ -68,7 +68,7 @@ which simplifies for [`AdditionGroupOperation`](@ref) to ``Dι_{$(_math(:G))}(g)
 """
 
 @doc "$(_doc_diff_inv_add)"
-diff_inv(G::LieGroup{𝔽,AdditionGroupOperation}, g, X) where {𝔽} = -X
+diff_inv(G::LieGroup{𝔽,AdditionGroupOperation}, g, X) where {𝔽}
```
mateuszbaran: Why did you remove the implementation?

kellertuer: I do not fully remember – this PR is soooo long by now, that I might have missed a few details along the way already. I also spent most of the last 2 weeks just working on the humungous test coverage – that also lead to me forgetting some of the details before.

I think that default implementation breaks some tests?

kellertuer:
Oh, noes, your last commit basically removed the amount of percentage in the code coverage that took me 3 hours of work yesterday (yes 2% of code coverage yesterday was 3 hours of work – yes I also feel suuuuuper slow with that).

src/group_operations/addition_operation.jl
kellertuer: Ah I think its ambiguities. Without that definition, this line is just  attaching a doc string without doing a function definition.

Feel free to fight ambiguities there, but I feel I did nothing else the last 6 weeks-or-so than fighting ambiguities and/or trying to get test coverage done. So after a few hours I am sometimes just tired and if all tests work without the “fancy generic default”, and with that fancy default I am having several special cases to define anyways – I by now (after 2 months of just working on this code-wise) get a bit tired.

mateuszbaran: I will check that then.

kellertuer: And sure – fancy defaults are super nice to have – but for me there is a tradeoff in time-consuming fights as well. I would prefer this PR to be of finite length at some point (that I do not see yet).

kellertuer: > I will check that then.

Good luck. I think I tried for a whole weekend or so.

mateuszbaran:
> Oh, noes, your last commit basically removed the amount of percentage in the code coverage that took me 3 hours of work yesterday (yes 2% of code coverage yesterday was 3 hours of work – yes I also feel suuuuuper slow with that).

That's weird, I haven't changed any logic anywhere.

kellertuer:
> That's weird, I haven't changed any logic anywhere.

You changed a few `Val{1}` of just `1`s. That was changing most of the submanifold-access-checks I think.

mateuszbaran:
If there is code that was covered with `1` but isn't with `Val(1)`, it most likely can be just removed? Tests are passing still.

src/interface.jl
```diff
@@ -83,33 +83,67 @@ Identity(::LieGroup{𝔽,O}) where {𝔽,O<:AbstractGroupOperation} = Identity{O
...
+"""
+    AbstractLieGroupPoint <: ManifoldsBase.AbstractManifoldPoint end
+
+An abstract type for a point on a [`LieGroup`](@ref).
+While an points and tangent vectors are usually kept untyped for flexibility,
+it might be necessary to distinguish different types of points, for example
+
+* for complicated representations that require a struct
+@ semantic verification
+* when there exist different representations
```
mateuszbaran: What does this `@` mean here?

src/group_operations/addition_operation.jl
mateuszbaran: I've just added it back and it just works.

src/interface.jl
kellertuer: Ah that is just a typo and should be a `*` (both @ and * are the same key on the Norwegian Keyboard, * being shift+@).

kellertuer:
> If there is code that was covered with `1` but isn't with `Val(1)`, it most likely can be just removed? Tests are passing still.

Maybe? I usually feel a bit not-so-clever when it comes to these dispatch ways with ? Val` where they are needed and where not.

src/group_operations/addition_operation.jl
kellertuer: Hm, then I am not sure why I removed it. This PR had a lot of fundamental changes, some even back and forth, on the interface – maybe I also removed it, when we still had the Lie groups exp's on `Identity` or so. But good that that works again, it probably makes a few definitions on the concrete addition groups irrelevant then.

test/groups/test_special_unitary_group.jl
```diff
@@ -0,0 +1,41 @@
...
+!(s in LOAD_PATH) && (push!(LOAD_PATH, s))
+using LieGroupsTestSuite
+
+@testset "Special Unitary Group" begin
+    G = SpecialUnitaryGroup(2)
+    #g1 = 1 / sqrt(2) * [1.0 1.0; -1.0 1.0]
+    #g2 = [0.0 -1.0; 1.0 0.0]
+    #g3 = [1.0 0.0; 0.0 1.0]
+    #X1, X2, X3 = [1.0 0.0; 0.0 0.0], [0.0 0.0; 0.0 1.0], [0.0 0.5; 0.5 0.0]
+    properties = Dict(
```
mateuszbaran: We should have some tests for different operations here I think, not just `show`.

kellertuer: Sure, as I wrote above, test coverage might not yet be fully complete, but I work on it literally every day to improve that a bit. Feel free to add tests as well.
Over the weekend I managed to cover some 170+ lines of code, but sure there is both about 100+ lines and a few semantic tests left.

src/interface.jl
```diff
@@ -83,33 +83,67 @@ Identity(::LieGroup{𝔽,O}) where {𝔽,O<:AbstractGroupOperation} = Identity{O
...
 @inline function ManifoldsBase._get_coordinates!(
-    G::LieGroup, Y, p, X, B::LieAlgebraOrthogonalBasis
+    G::LieGroup, c, X, B::DefaultLieAlgebraOrthogonalBasis
 )
-    return get_coordinates_lie!(G, Y, p, X, number_system(B))
+    return get_coordinates_lie!(G, c, X, B)
 end
-@inline function ManifoldsBase._get_vector(G::LieGroup, p, c, B::LieAlgebraOrthogonalBasis)
-    return get_vector_lie(G, p, c, number_system(B))
+@inline function ManifoldsBase._get_vector(
```
mateuszbaran: I think we don't have any tests for `get_vector` with `LieGroup` as the first argument here. I'm not sure if we need it when it works for `LieAlgebra` without the point already?

mateuszbaran:
I haven't checked the logic for the semidirect product yet but the general design looks fine (maybe up to the `get_vector`/`get_coordinates` discussion I've started above). I think this PR is on the right track.

kellertuer:
Thanks for that feedback, design-wise getting semidirect generically correct took the longest time (code wise, the dispatch with Identity – and overall for sure testing until now took the longest time).

src/interface.jl
kellertuer: Yeah, that is probably now a bit of a discussion, where `(G, p, ...` should / does exist, when `(𝔤,...` also exists.

For now I went for
a) if this is really and only ever Lie algebra stuff, only the second exists
b) if there might be a good reason that the first exists and g not the identity is common in practice, the first exists as well

Having in mind the name of the Basis type I agree, that this variant here is not of case b) and hence should be removed.

src/groups/orthogonal_group.jl
```diff
@@ -0,0 +1,528 @@
...
+
+function _get_vector_lie_On!(X, c)
+    n = size(X, 1)
+    @inbounds begin
+        X[1, 2] = -c[3]
+        X[2, 1] = c[3]
+        X[1, 3] = c[2]
+        X[3, 1] = -c[2]
+        X[2, 3] = -c[1]
+        X[3, 2] = c[1]
```
mateuszbaran, line 353: It looks like you removed zeroing the diagonal when copying this from Manifolds.jl? That's most likely the cause of the recent failure on CI but maybe you had an idea behind that?

kellertuer: If I had an idea, same as in one of the last posts, I forgot about it by now :D (this PR is really a bit long time wise by now) so if that caused CI to fail it should be re-added, if they make CI fail now.
I actually can not even find where that is in Manifolds.jl...

mateuszbaran: I've added setting zeros back

kellertuer:
Now just three very specific test cases are missing

* ~~`adjoint` (and its allocation) for a specific wrapper type (which we do not have but we could do a dummy type~~
* ~~[ ] `conjugate_jacobian` (allocation and call to mutating variant) – I have to check where to implement one best~~
* [x] `inv` (for left/right) and `identity_element` in their generic implementation for semi direct products; basically the automatic implementation that is available for arbitrary such Lie groups – I have to check whether we already have a nice one (this is not a specific one like SE(n) with its own implementation)

But is is also only 28 lines left, so the end is in sight, though for neither of the three I have a good idea how to write a test yet.

_edit_: Just before I wanted to leave I had an idea for the third, which in principle now exists, but there is a bug in my `inv`s (maybe some confusion with me and the theory again). Will check back.

test/groups/test_special_unitary_group.jl
kellertuer: Hm, this was a good note, but for the last 30 minutes I just see errors increasing again. So both this and the unitary group probably still need quiiiiiiiiiiite some work, probably especially allocation magic is off often and allocates real stuff when it should allocate complex.

kellertuer: Ok, the main things not working are `hat`/ `vee` because we do not have any basis implemented in the complex case for general unitary (especially no ONB that the LieAlgebraBasis wants to fall back to) – and `rand` also due to missing support on Manifolds. Is that ok to first fix those over there and come back here on a later PR?
_edit:_ With `hat`/`vee` the same on unitary, `rand` however works there.

kellertuer:
Final update for today (got a bit carried away)

* For adjoint I would wait for a manifold that wraps and has a nice adjoint – for example the `ValidationLieGroup` planned soon
* similarly for the conjugate jacobian I would wait for a Lie group having this.

Since we at least make code cov already happy with only 8 new lines – we had 6 before, but this PR is larger than all previous code... maybe that is ok?

mateuszbaran:
Nice! I think we can skip `adjoint` for now but I'd prefer to finish `jacobian_conjugate`. We could just have a default implementation based on diff_conjugate, similarly to what we have for `adjoint_matrix` in Manifolds.jl. Something more or less like this:
```julia
function jacobian_conjugate!(
    G::LieGroup, J, g, h; B::AbstractBasis=DefaultLieAlgebraOrthogonalBasis()
)
    Bb = get_basis(G, h, B)
    Vs = get_vectors(G, h, Bb)
    for i in eachindex(Vs)
        get_coordinates!(G, view(J, :, i), g, diff_conjugate(G, g, h, Vs[i]), B)
    end
    return J
end
```
This doesn't work because there are some issues with bases which we also might want to resolve here. I'm not 100% sure if I didn't mix `g` and `h` here so this also needs to be checked after bases work.

docs/src/tutorials/transition.md
```diff
@@ -23,38 +23,38 @@ The list is alphabetical, but first lists types, then functions
...
 | `apply_diff_group` | [`diff_group_apply`](@ref) | modifiers (diff/group) come first, consistent with [`ManifoldsDiff.jl`](https://juliamanifolds.github.io/ManifoldDiff.jl/stable/) |
 | | [`conjugate`](@ref), [`diff_conjugate`](@ref) | a new function to model ``c_g: \mathcal G → \mathcal G`` given by ``c_g(h) = g∘h∘g^{-1}`` |
-| `exp(G, g, X)` | `exp(`[`base_manifold`](@ref base_manifold(G::LieGroup))`(G), g, X)` | the previous defaults whenever not agreeing with the invariant one can now be accessed on the internal manifold |
-| `exp_inv(G, g, X)` | [`exp`](@ref exp(G::LieGroup, g, X, t::Number))`(G, g, X)`  | the exponential map invariant to the group operation is the default on Lie groups here |
-| `exp_lie(G, X)` | [`exp`](@ref exp(G::LieGroup, e::Identity, X, t::Number))`(G, `[`Identity`](@ref)`(G), X)` | the (matrix) exponential is now the one at the [`Identity`](@ref)`(G)`, since there it agrees with the invariant one |
+| `differential_exp_argument_lie_approx` | - | discontinued due to name change. Though available in `ManifoldDiff.jl` for `GroupManifolds`, that will move to `differential_exp_argument` instead, since `exp_lie` changed to now just `exp`. |
```
mateuszbaran:
```suggestion
| `differential_exp_argument_lie_approx` | - | Scheduled for update and renaming. Though available in `ManifoldDiff.jl` for `GroupManifolds`, that will move to `differential_exp_argument_approx` instead, since `exp_lie` changed to now just `exp`. |
```

kellertuer:
Hm, I do not yet understand this – neither your proposed code, nor why we have “some issues with bases” (and where), so I have to first read up on all that.

mateuszbaran:
This is calculating Jacobian by separately passing each basis vector through the `diff` function. I think this should be a standard fallback pattern for Jacobians in general?

mateuszbaran:
Anyway, you'll see the issue when you try running this code.

kellertuer:
I mainly mean all the nitty gritty details, that the differential could be seen as a map between tangent spaces but for us the tangent spaces are all identified with the Lie group – but does that then involve `divv_inv_left_compose` or not or what do we expect the differential to be given in...?

The docs on the Jacobian are still a bit short on that so I have to track that down first before I can say something about your code – and I also would need a concrete example to run this on, otherwise I can not just run that code.

mateuszbaran:
Right, it might not be that straightforward and isn't crucial to this PR. I will just open an issue about `jacobian_conjugate`.

kellertuer:
thanks – I do agree that that should be provided, but I would need probably a day or two to get the notation in the docs right, understand the form and implement it correctly.

I can do that after the ValidationLieGroup (which is my next plan after this PR)

mateuszbaran:
There is no rush, I'd just like to have `jacobian_conjugate` working here before groups are removed from Manifolds.jl.

kellertuer:
I do fully agree on that. Then it is very good to have an issue open about that.

test/groups/test_special_unitary_group.jl
mateuszbaran: I think we can skip `hat` and `vee` for now.

mateuszbaran (approved):
I think this is good enough now. I've opened one issue to check some aspects of SE(N) which are handled by Manifolds.jl but probably not here but it can be handled later (though before removing groups from Manifolds.jl). We should also note that quaternionic unitary needs to be handled (though just a point on the roadmap is enough).

kellertuer:
SE(n) should be more generic here, because it works on (R,t), (R,t) and the matrix containing R t in homogeneous coordinates – but sure, performance is not my main area of expertise.

mateuszbaran:
I know, and that's good. I mean primarily performance, support for AD and handling different numeric types.

kellertuer:
Yes, very good to keep these in mind and have those aspects covered as well. Thanks for having that in mind.

I will merge this later today after my meetings and lectures then.

PR #18, opened 2024-12-19
Improve docs of group exp and log, mention left-invariance

kellertuer:
This resolves #16. 
cc @olivierverdier

Issue #16, opened 2024-11-20
Improve `exp` and `log` documentation strings


kellertuer:
Currently it is not mentioned that

* `X` is from the Lie algebra, as we use that as the default representation of tangent vectors
* we have the left-invariant metric in mind

(cc @olivierverdier )

olivierverdier:
Just being pedantic: this has nothing to do with a metric, it is rather the exponential associated to the canonical group connection (which need not be metric). There are always two exactly equivalent ways to compute the exponential (here $`\exp_1`$ is the Lie exponential):
- $`\exp_p(X) = p \exp_{1}(p^{-1}X)`$
- $`\exp_p(X) = \exp_{1}(Xp^{-1}) p`$

With the convention of storing a vector field $`X`$ at $`p`$ as $`X = pY`$, the first formulation is much better, since in terms of $`Y`$ you have (just caring about the storage): $`\exp_p(Y) \equiv \exp_1(Y)`$.

kellertuer:
Ah yes, I sometimes am not exact enough with metric and connection. Yes our default convention would be pY and just store Y; but we could also reflect that in tangent-vector-types as also discussed yesterday.

olivierverdier:
Same issue with the `log` docstring. If $`\log_1`$ is the Lie logarithm, then the group log fulfills either of those
- $`\log_p(q) = p\log_1(p^{-1}q)`$
- $`\log_p(q) = \log_1(qp^{-1})p`$

And again, with the left-invariant storage convention, the first formulation is better, since in the end you just compute $`\log_p(Y) \equiv \log_1(p^{-1}q)`$.

kellertuer:
Sure. I'll add the title here right away.

PR #19, opened 2025-01-01
add Heisenberg group

mateuszbaran:
Do we have it documented somewhere what orthogonality means on a Lie group?


kellertuer:
Can you elaborate what orthogonality you mean?

src/groups/heisenberg_group.jl
```diff
@@ -0,0 +1,369 @@
...
+\begin{bmatrix} 1 & \mathbf{a} & c \\
+\mathbf{0} & I_n & \mathbf{b} \\
+0 & \mathbf{0} & 1 \end{bmatrix}
+```
+
+where ``I_n`` is the ``n×n`` unit matrix, ``\mathbf{a}`` is a row vector of length ``n``,
+``\mathbf{b}`` is a column vector of length ``n`` and ``c`` is a real number.
+The group operation is matrix multiplication.
+
+The left-invariant metric on the manifold is used.
```
kellertuer, line 17: In #18 this is introduced as the global default and explained a bit more. I think we could leave it out here, then?

src/groups/heisenberg_group.jl
```diff
@@ -0,0 +1,369 @@
...
+}
+
+function HeisenbergGroup(n::Int; parameter::Symbol=:type)
+    Hm = Manifolds.HeisenbergMatrices(n; parameter=parameter)
+    return HeisenbergGroup{typeof(Hm).parameters...}(
+        Hm, MatrixMultiplicationGroupOperation()
+    )
+end
+
+function active_traits(f, ::HeisenbergGroup, args...)
```
kellertuer: `LieGroup` is not a decorator manifold (on purpose) so, this has zero effect- all `GroupManifold` things are deprecated (soon) anyways.

Yeah, I should write a tutorial after finishing SE

test/LieGroupsTestSuite.jl/LieGroupsTestSuite.jl
```diff
@@ -398,15 +398,15 @@ function test_exp_log(
                 exp!(G, k2, e, X)
                 @test isapprox(G, k1, k2)
             end
-            @test is_point(G, k1)
+            @test is_point(G, k1; error=:warn)
```
kellertuer: I would maybe even error these, a failing test would print an error already, this error would give more information?

mateuszbaran:
I mean, we need a metric to call a basis orthonormal or orthogonal. When the exp/log of that metric don't match exp/log we have for that group, it might be a bit weird.

test/groups/test_heisenberg_group.jl
```diff
@@ -0,0 +1,49 @@
...
+    [0.0 4.0 -3.0; 0.0 0.0 3.0; 0.0 0.0 0.0],
+    [0.0 -2.0 1.0; 0.0 0.0 1.1; 0.0 0.0 0.0]
+
+    properties = Dict(
+        :Name => "Heisenberg group",
+        :Points => [g1, g2, g3],
+        :Vectors => [X1, X2, X3],
+        :Rng => Random.MersenneTwister(),
+        :Functions => [
+            compose,
```
kellertuer, line 26:
```suggestion
            adjoint,
            compose,
```

also here, my tutorial will also cover tests and which functions to usually test ;)

test/LieGroupsTestSuite.jl/LieGroupsTestSuite.jl
mateuszbaran: A failing test doesn't tell why the point was incorrect without this.

kellertuer:
> I mean, we need a metric to call a basis orthonormal or orthogonal. When the exp/log of that metric don't match exp/log we have for that group, it might be a bit weird.

On the _Lie group_ there is no metric, that is why I was confused; but you refer to the Lie algebra? That does come with an inner product anyways since it is a tangent space.
What is the actual point where this is currently a problem? If it is the `DefaultLieAlgebraOrthogonalBasis` I am fine renaming that to a name without `Orthogonal`

src/groups/heisenberg_group.jl
mateuszbaran: Not all groups use a left-invariant metric though? So it doesn't seem redundant?

test/LieGroupsTestSuite.jl/LieGroupsTestSuite.jl
kellertuer: until now I set that to `error=:error`, if you prefer to warn only (and the failing test to error), we could change that everywhere, sure.

mateuszbaran:
Hm, it looks like I forgot again how orthogonality works in LieGroups.jl.

test/LieGroupsTestSuite.jl/LieGroupsTestSuite.jl
mateuszbaran: I can change it to error.

src/groups/heisenberg_group.jl
kellertuer: Well, at some point we have to decide on something. After long discussions with you and Olivier, I thought we would make “left invariant metric” the global default. That is what #16 was about and what #18 tries to specify in the docs.

Sure that does mean for those that do use that as their default it has to be mentioned. From what I experience until now, the overal l default _always_ seems to be left-inv, so that this default seems reasonable to me.
If not, feel free to close #18 and do it better, I am just trying my best here. I do feel mentioning the global default again _is_ redundant, since it is the global default.

mateuszbaran: A big point in discussions with Oliver was about how the left-invariant metric is a bad choice for SE(n). We settled on left-invariant tangent vector representation, but it's a much more restricted thing than the whole metric structure.

mateuszbaran: In particular, SE is supposed to have no metric IIRC. So having `inner` or `norm` defined can be misleading.

kellertuer: Now I am confused. I read about SE most of the last few days, but now I feel, maybe I do not understand anything of it. Then maybe someone else should both implement that and improve #18? I slowly feel I am indeed not the right person to do this package at all the more I work on it.

In the current form _all_ Algebras have an inner since _all_ of them are tangent vectors. 
SE is a Lie group and hence a manifold and hence has no inner, so your second part also super-confuses me.

kellertuer: And – should I then just close #18 since it seems I still do not understand the real reason for #16? Olivier again just said I am wrong and I should fix it. I tried my best, but it seems, yet again, I am wrong.

maybe it is really best I give all this a few weeks of a break. Whenever I feel I do the smallest step forwards it ends with me feeling doing at least 19 steps back.

#17 can be shortened to leaving out SE and then someone else does that. It should provide all tools to be able to implement SE.

mateuszbaran: On Riemannian manifolds parallel transport preserves `inner` between vectors, but AFAIK on SE the PT we are supposed to have does not. Maybe I put too much emphasis on this invariance but for some reason I keep getting anxious about having a non-invariant inner. I'm not sure at this point. We should at least clearly document that.

mateuszbaran: SE is supposed to be the "typical case", not the odd one. It's the Lie groups with a bi-invariant Riemannian metrics that are special cases. What's wrong with #18?

kellertuer: Currently that PR says “By default and if not mentioned otherwise, the typical case are left-invariant metrics”..
Because all that I implemented for the last few months assumes exactly that.

Now we are back to that this is not the default. Then we have to store the type of metric again somewhere. Then you can rework 90% of this package and reintroduce another left/right/jing/jang parameter type somewhere to store that.
That is what‘s wrong.

kellertuer: For PT: I have a bit of an idea about PT on product manifolds and Lie groups.
For Semidirect ones, I do not even have a clue about exp in general even less PT.
If the PT is the once corresponding to the metric, it _by definition_(!) preserves inner products, otherwise it is not the PT belonging to that metric.

And sure there are probably several different inners
* $T_eM$ has an inner with respect to the Riemannian metric
* the Lie algebra might bring another one, but it is also its own type now

my main problem is, that this discussion now really destroys most of the modelling I did for the last months. So if we do not assume by default left-invariance (like we also assume a default metric on manifolds) and document other Lie groups with theirs as default when we get to them.
Then we have to store the metric invariance somewhere.
But then we have to rework the whole package.

I will have to first start a semester on Monday, so such a large thing, reworking all the work I have done for this package until now – I can not do while starting a semester with a large lecture,.
I had hopes we reached some kind of maturity in this package by now, but maybe this concludes, we did not.

kellertuer: Let me know what you think, if left invariance is not the default, then I stop working on this package until we have an issue open and clarified how this should be handled in the package and I continue with any coding after that.

I would prefer to not code too much that gets thrown away, and my feeling for now is, that all current code base here is – due to this discussion – useless.

mateuszbaran: I think we maybe should have a short meeting about that, it would be easier than writing another long post about relationship between Lie groups and Riemannian metrics. I think your work here is very much not useless. 

> Currently that PR says “By default and if not mentioned otherwise, the typical case are left-invariant metrics”..

I don't see that there?

> I would prefer to not code too much that gets thrown away, and my feeling for now is, that all current code base here is – due to this discussion – useless.

I had this PT/inner in mind when reviewing your work here, at least sometimes, and there is nothing particularly wrong with the design. There is just standard Lie group stuff that we have to be careful with bases. Nothing can change that, it's an inherent issue.

kellertuer: Currently _all_ of LieGroups.jl – probably far too implicitly – assumes left-invariant vector fields / metrics on the Lie algebra.

Yeah a meeting is probably best. I am not sure the current design covers your current worries. In like not at all. Especially if we have Lie groups where we have _different_ invariant vector fields for the same Lie algebra we want to implement something for.

If a single Lie algebra uses right invariant vector fields consistenty that might be fine, not sure.

Sure nothing can change that we have to be careful, still, I feel that I am not versed enough to continue this work. Before this thread I was slowly reading up on SE to maybe start implementing that soon. Now that is has to be (?) w.r.t. right invariant metric (no clue why, saw that nowhere). I am back to the start.

src/groups/heisenberg_group.jl
mateuszbaran: Fixed.

kellertuer (commented):
Up to one small remark this looks good. Thanks!

src/groups/heisenberg_group.jl
```diff
@@ -0,0 +1,200 @@
...
+    a_view = _heisenberg_a_view(M, X)
+    b_view = _heisenberg_b_view(M, X)
+    q[1, 2:(n + 1)] .= a_view
+    q[2:(n + 1), n + 2] .= b_view
+    q[1, n + 2] = X[1, n + 2] + dot(a_view, b_view) / 2
+    return q
+end
+
+@doc raw"""
+    exp(M::HeisenbergGroup, p, X)
```
kellertuer: Can we maybe use G (not M) for the group, and g,h (not p,q) ? That avoids name clashes when referring to the (underlying) manifold

(sorry this should have been included in the review before, which GitHub decided to leave empty, and delete what I already typed here before. no clue why)

mateuszbaran: Sure, I've updated the naming.

kellertuer (approved):

Issue #20, opened 2025-02-19
Implement generic `jacobian_conjugate`


mateuszbaran:
It should be doable based on `diff_conjugate`, see https://github.com/JuliaManifolds/LieGroups.jl/pull/17#issuecomment-2666778695 , similarly to `affine_matrix` in Manifolds.jl.

PR #22, opened 2025-02-23
Add a ValidationLieGroup

kellertuer:
📋

This PR aims to provide the analogue of a [`ValidationManifold`](https://juliamanifolds.github.io/ManifoldsBase.jl/stable/manifolds/#A-manifold-for-validation) for Lie groups. Internally one can (but does not have to) add such a validation also for the inner manifold functions.

* [x] implement the Lie groups interface
* [x] 📚 adapt docs
* [x] resolve ambiguties 
* [x] test coverage

### DIFF
--- a/src/groups/validation_group.jl
+++ b/src/groups/validation_group.jl
@@ -348,6 +348,31 @@ function ManifoldsBase.exp!(G::ValidationLieGroup, h, g, X; kwargs...)
     return g
 end
 
+function hat(
+    𝔤::LieAlgebra{𝔽,O,<:ValidationLieGroup}, c; kwargs...
+) where {𝔽,O<:AbstractGroupOperation}
+    G = base_lie_group(𝔤).lie_group
+    X = hat(LieAlgebra(G), c)
+    is_point(𝔤, X; widthin=hat, context=(:Output,), kwargs...)
+    return ValidationLieAlgebraTangentVector(X)
+end
+function hat(
+    𝔤::LieAlgebra{𝔽,O,<:ValidationLieGroup}, c, T::Type; kwargs...
+) where {𝔽,O<:AbstractGroupOperation}
+    G = base_lie_group(𝔤).lie_group
+    X = hat(LieAlgebra(G), c, T)
+    is_point(𝔤, X; widthin=hat, context=(:Output,), kwargs...)
+    return ValidationLieAlgebraTangentVector(X)
+end
+function hat!(
+    𝔤::LieAlgebra{𝔽,O,<:ValidationLieGroup}, X, c; kwargs...
+) where {𝔽,O<:AbstractGroupOperation}
+    G = base_lie_group(𝔤).lie_group
+    hat!(LieAlgebra(G), internal_value(X), c)
+    is_point(𝔤, X; widthin=hat, context=(:Output,), kwargs...)
+    return X
+end
+
 function identity_element(G::ValidationLieGroup; kwargs...)
     g = identity_element(G.lie_group)
     is_point(G, g; widthin=identity_element, context=(:Output,), kwargs...)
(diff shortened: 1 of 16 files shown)


kellertuer:
This basically just needs a few functions to be defined for the `LieAlgebra` of a `ValidatinLieGroup` to handle wrapped points a bit – and a careful check, whether or not `Identity(G)` should be wrapped as `ValidationMPoint(Identity(G))` or not.

kellertuer:
Decided again st wrapping identity, since that causes too much strange side effects on the system.

kellertuer:
I consider this ready for review, since the only thing that should be missing by now is a few test coverage corner cases.

kellertuer:
The only lines missing are the ones that validate `jacobian_conjugate`, which we can do a test for as soon as we have the default from #20.

So this PR is complete.

src/groups/validation_group.jl
```diff
@@ -0,0 +1,760 @@
...
+* `:All`: disable all checks
+* `:Point`: checks for points
+* `:Algebra`: checks related to the [`LieAlgebra`](@ref)
+* `:Output`: checks for output
+* `:Input`: checks for input variables
+
+# Fields
+
+* `lie_group::L` the [`AbstractLieGroup`](@ref) to be decorated
+* ``
```
mateuszbaran: Other fields seem to be missing, though we could probably just link to `ValidationManifold` for their description`?

src/groups/validation_group.jl
```diff
@@ -0,0 +1,760 @@
...
+    error=G.mode,
+    within::Union{Nothing,<:Function}=nothing,
+    context::Union{NTuple{N,Symbol} where N}=NTuple{0,Symbol}(),
+)
+    !_vLc(G, within, context) && return nothing
+    (error === :error) && (throw(ErrorException(str)))
+    (error === :warn) && (@warn str)
+    (error === :info) && (@info str)
+    return nothing
+end
```
mateuszbaran, line 167: Could you maybe reuse `_msg` from ManifoldsBase? If you extend `_vMc` instead of defining `_vLc` it should work fine I think?

src/groups/validation_group.jl
kellertuer: Ah I forgot to add them, yes it is basically the same fields (again) as for the Validation one, and I would even try to keep both parallel – since that way one can pass down the same things and validate on M as well – or not – if one wants to.

mateuszbaran:
I have only these two comments, everything else looks fine.

src/groups/validation_group.jl
kellertuer: there was some reason why I had to rewrite `_vMc` to `_vLc`, the `_msg` I then rewrote to not depend on ManifoldsBase internals – though that is of course more of a philosophical point – that way we can change the internals of ManifoldsBase without breaking stuff here.

src/groups/validation_group.jl
kellertuer: Fixed it :)

kellertuer:
It was also mainly a challenge to work through all the functions and not necessarily challenging from the intellectual side ;)

src/groups/validation_group.jl
mateuszbaran: I see. I would prefer less duplication but we can keep it this way too.

mateuszbaran (approved):

src/groups/validation_group.jl
kellertuer: I do see positive points in both approaches. Mine is a bit more independent of internal things of another package – yours avoids a few lines of code duplication. But thanks that we can leave it like this for now.

If we come across a good reason to have these few lines not double, we can also change it later.

kellertuer:
Nice! Then let's merge this one first, I can care about the missing code coverage in the jacobian PR then.

PR #23, opened 2025-03-01
Implement a default for `jacobian_conjugate`

kellertuer:
This solves #20.

### DIFF
--- a/src/groups/product_group.jl
+++ b/src/groups/product_group.jl
@@ -344,7 +344,7 @@ function Base.show(
 end
 
 function get_coordinates_lie!(
-    Pr𝔤::LieAlgebra{𝔽,Op,LieGroup{𝔽,Op,M}}, c, X
+    Pr𝔤::LieAlgebra{𝔽,Op,LieGroup{𝔽,Op,M}}, c, X, B::DefaultLieAlgebraOrthogonalBasis
 ) where {𝔽,Op<:AbstractProductGroupOperation,M<:ProductManifold}
     PrG = Pr𝔤.manifold
     PrM = PrG.manifold
--- a/src/groups/product_group.jl
+++ b/src/groups/product_group.jl
@@ -344,7 +344,7 @@ function Base.show(
 end
 
 function get_coordinates_lie!(
-    Pr𝔤::LieAlgebra{𝔽,Op,LieGroup{𝔽,Op,M}}, c, X, B::DefaultLieAlgebraOrthogonalBasis
+    Pr𝔤::LieAlgebra{𝔽,Op,LieGroup{𝔽,Op,M}}, c, X
 ) where {𝔽,Op<:AbstractProductGroupOperation,M<:ProductManifold}
     PrG = Pr𝔤.manifold
     PrM = PrG.manifold
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -21,7 +21,7 @@ at the [`identity_element`](@ref) on the [`base_manifold](@ref base_manifold(::A
 
 !!! note
     In order to implement the corresponding [`get_coordinates`](@ref) and [`get_vector`](@ref) functions,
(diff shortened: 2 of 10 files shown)


src/groups/product_group.jl
```diff
@@ -344,7 +344,7 @@ function Base.show(
 end
 
 function get_coordinates_lie!(
-    Pr𝔤::LieAlgebra{𝔽,Op,LieGroup{𝔽,Op,M}}, c, X, B::DefaultLieAlgebraOrthogonalBasis
+    Pr𝔤::LieAlgebra{𝔽,Op,LieGroup{𝔽,Op,M}}, c, X
```
mateuszbaran: Why isn't ` B` needed here?

kellertuer: `get_coordinates_lie` is already the specific (Layer3) function for the `DefaultLieAlgebraOrthogonalBasis`, so this function does not need a basis any longer.

mateuszbaran: There are variants of `DefaultLieAlgebraOrthogonalBasis` with real and complex coefficients and this information needs to be propagated here somehow (and I don't like passing just the number system, it's too much hassle).

mateuszbaran: Anyway, `B` is even used inside so this function can't work without it.

kellertuer: Hm, then this is overall quite inconsistent throughout LieGroups for now, I think I never passed (or dispatched on) basis in this Layer3 function.
But sure for other Layer-3-cases we do that, though most often with dispatching on the number system type I think.

mateuszbaran: Yes, and IMO it was a bad idea to not pass the basis to layer-3 functions. I think this is a good time to stop doing it. I'm planning to redo ManifoldsBase dispatch accordingly when I have some spare time.

src/interface.jl
```diff
@@ -645,9 +651,27 @@ end
...
+    n = number_of_coordinates(base_manifold(G), B)
+    𝔤 = LieAlgebra(G)
+    c = zeros(eltype(J), n)
+    for j in 1:n
+        c .= 0
+        c[j] = 1
+        get_vector!(𝔤, X, c, B)         # store ``X_j`` in X
+        diff_conjugate!(G, X, g, h, X) # compute the differential in-place
+        get_coordinates!(𝔤, c, X, B)   # compute its coordinates in B
+        J[:, j] .= c                    # store the result in J
```
```suggestion
        get_coordinates!(𝔤, view(J, :, j), X, B)   # compute its coordinates in B and store the result in J
```
This can work in one step

src/groups/product_group.jl
kellertuer: I do not have a strong opinion on that, besides that it should be unified to some extend. Sure we can read it here, and even more if it is used inside, I seem to have missed it (and interesting that the tests work?) – and rework ManifoldsBase internally sounds also good then.

Will do that together with the code coverage for ValidationManifold after merging master back here – but not today. Tomorrow or Thursday probably.

mateuszbaran: I did not have a strong opinion on that when it was introduced but I do have a strong opinion now after working with the current design a bit.

mateuszbaran:
Everything else looks good to me.

src/groups/product_group.jl
kellertuer: Sure, sometimes it needs some time and using stuff to see what is better.

kellertuer: But yeah, I merged master with the `ValidationLieGroup` but will add `B` than back in tomorrow and check that this is done everywhere for `get_vector_lie` and `get_coordinates_lie` then.

kellertuer:
Fixed the issue, was good, now all is unified – also brought validation jacobian_conjugate into the tests, so code cov also looks good. I think there is no need for an additional entry in the changelog, all is already mentioned.

mateuszbaran (approved):

PR #24, opened 2025-03-01
A get-started tutorial

kellertuer:
This is a start of a tutorial that aims to combine the former two

* introduction to `GroupManifolds` at https://juliamanifolds.github.io/Manifolds.jl/stable/tutorials/groups/
* the stale PR https://github.com/JuliaManifolds/Manifolds.jl/pull/366

For now this is just a short start, but I will continue the already prepared roadmap in the quarto notebook.


kellertuer:
@Affie @dehann if you have a bit of time, this aims to replace your started tutorial from a few years back – SE(3) I still have to write.
The focus here is much more on Lie groups than in the two linked ones. Since one of them was also a discussion with you, it would be great to have some feedback from you here as well.

kellertuer:
All text and code are written now, maybe still a bit technical, but all functions I wanted to mention and introduce are now included.

See https://juliamanifolds.github.io/LieGroups.jl/previews/PR24/tutorials/getstarted/ for the rendered version.

tutorials/getstarted.qmd
```diff
@@ -1,3 +1,312 @@
...
+```math
+R_α = \begin{pmatrix} \cos(α),& -\sin(α)\\ \sin(α) & \cos(α) \end{pmatrix},\qquad α ∈ ℝ,
+```
+
+where we already represent all possible rotations when restricting $α$ to $[0,2π]$.
+Adding two matrices $R_α + R_β$ does not yield a valid rotation matrix again, but we
+can have these matrices in mind as being “locally like the real line”, since changing the rotation angle
+slightly yields rotation matrices that are “close by” in the sense that they perform nearly the same rotation
+when applied to a point $x ∈ ℝ^2$ by computing $R_αx$.
+Matrix multiplication $R_β = R_{α_1}*R_{α_1}$ does yield a new rotation and also fulfils all other
```
mateuszbaran:
```suggestion
Matrix multiplication $R_β = R_{α_1} R_{α_2}$ does yield a new rotation and also fulfills all other
```
I'm not a fan of using `*` for multiplication in math.

tutorials/getstarted.qmd
```diff
@@ -1,3 +1,312 @@
...
+`LieGroups.jl` uses the manifolds defined in [`Manifolds.jl`](https://juliamanifolds.github.io/Manifolds.jl/stable/). For more details on that, see also the [introductory tutorial](https://juliamanifolds.github.io/Manifolds.jl/stable/tutorials/getstarted/) there.
+A Lie group is usually written as a tuple $(ℳ,∘)$ os the manifold and the group operation thereon.
+
+For more theoretical background, see for example [HilgertNeeb:2012](@cite), especially their Chapter 9.
+
+## Rotations on the plane $ℝ^2$ and in $ℝ^3$.
+
+This first part considers rotations in the plane. These can be represented by rotation matrices
+```math
+R_α = \begin{pmatrix} \cos(α),& -\sin(α)\\ \sin(α) & \cos(α) \end{pmatrix},\qquad α ∈ ℝ,
```
mateuszbaran:
```suggestion
R_α = \begin{pmatrix} \cos(α)& -\sin(α)\\ \sin(α) & \cos(α) \end{pmatrix},\qquad α ∈ ℝ,
```

tutorials/getstarted.qmd
```diff
@@ -1,3 +1,312 @@
...
 
-In this tutorial, we aim to present how to get started with `LieGroups.jl`.
\ No newline at end of file
+## Introduction
+
+This tutorial introduces both a few basics of [Lie groups](https://en.wikipedia.org/wiki/Lie_group) as well as how to use [LieGroups.jl](https://juliamanifolds.github.io/LieGroups.jl/stable/). The Lie groups we consider are the rotations in the plane,
+or [special orthogonal group](../groups/special_orthogonal_group.md) as well as rigid body motions,
+or in other words the [special euclidean group](../groups/special_euclidean_group.md).
+
+In a nutshell a Lie group $𝒢$ is a [manifold](https://en.wikipedia.org/wiki/Manifold) $ℳ$ that also is a group a group operation.
```
mateuszbaran:
```suggestion
In a nutshell a Lie group $𝒢$ is a [manifold](https://en.wikipedia.org/wiki/Manifold) $ℳ$ that also is a group with a group operation.
```

tutorials/getstarted.qmd
```diff
@@ -1,3 +1,312 @@
...
+A manifold informally be described as a set that “locally looks like a Euclidean space”
```
mateuszbaran:
```suggestion
A manifold can informally be described as a set that “locally looks like a Euclidean space”
```

tutorials/getstarted.qmd
```diff
@@ -1,3 +1,312 @@
...
+and has some way to measure angles at every point, to be precise an inner product, that is allowed to (smoothly)
+depend on the point we are at.
```
mateuszbaran: This isn't necessary for Lie groups, so I'd suggest skipping the inner product here.

tutorials/getstarted.qmd
```diff
@@ -1,3 +1,312 @@
...
+A [group](https://en.wikipedia.org/wiki/Group_(mathematics)) means that on the set there also
+exists an operation, within this package usually denoted by $∘: 𝒢 × 𝒢 → 𝒢$$,
```
mateuszbaran:
```suggestion
exists an operation, within this package usually denoted by $∘: 𝒢 × 𝒢 → 𝒢$,
```

tutorials/getstarted.qmd
```diff
@@ -1,3 +1,312 @@
...
+`LieGroups.jl` uses the manifolds defined in [`Manifolds.jl`](https://juliamanifolds.github.io/Manifolds.jl/stable/). For more details on that, see also the [introductory tutorial](https://juliamanifolds.github.io/Manifolds.jl/stable/tutorials/getstarted/) there.
+A Lie group is usually written as a tuple $(ℳ,∘)$ os the manifold and the group operation thereon.
```
mateuszbaran:
```suggestion
A Lie group is usually written as a tuple $(ℳ,∘)$ with the manifold and the group operation thereon.
```

tutorials/getstarted.qmd
```diff
@@ -1,3 +1,312 @@
...
+R_α = \begin{pmatrix} \cos(α),& -\sin(α)\\ \sin(α) & \cos(α) \end{pmatrix},\qquad α ∈ ℝ,
+```
+
+where we already represent all possible rotations when restricting $α$ to $[0,2π]$.
+Adding two matrices $R_α + R_β$ does not yield a valid rotation matrix again, but we
+can have these matrices in mind as being “locally like the real line”, since changing the rotation angle
+slightly yields rotation matrices that are “close by” in the sense that they perform nearly the same rotation
+when applied to a point $x ∈ ℝ^2$ by computing $R_αx$.
+Matrix multiplication $R_β = R_{α_1}*R_{α_1}$ does yield a new rotation and also fulfils all other
+properties to yield a group operation. Using trigonometric identities we observe that $β = α_1 + α_2$.
```
mateuszbaran:
```suggestion
properties to yield a group operation. Using trigonometric identities we observe that $β = α_1 + α_2 \pmod{2\pi}$.
```

tutorials/getstarted.qmd
```diff
@@ -1,3 +1,312 @@
...
+Besides one special rotation in the following code for completeness, a prominent, at first glace
+maybe a bit “dull” point on this Lie group is the identity matrix `e`.
+
+```{julia}
+SO3 = SpecialOrthogonalGroup(3)
+g = [cos(π/3) -sin(π/3) 0.0; sin(π/3) cos(π/3) 0.0; 0.0 0.0 1.0]* [cos(π/4) 0.0 -sin(π/4); 0.0 1.0 0.0; sin(π/4) 0.0 cos(π/4)]
+e = Matrix{Float64}(I,3,3)
+g
+```
+The element `g` can be seen as a rotation by $π/3$ in the x-y-plane combines with a rotation by $π/4$ in the x-z plane.
```
mateuszbaran:
```suggestion
The element `g` can be seen as a rotation by $π/3$ in the x-y-plane combined with a rotation by $π/4$ in the x-z plane.
```

tutorials/getstarted.qmd
```diff
@@ -1,3 +1,312 @@
...
+## The Lie algebra, and the Lie group exponential function.
+
+For the following investigations, we consider the special orthogonal group $\mathrm{SO}(3)$,
+that is rotations in a 3-dimensional space.
+Besides one special rotation in the following code for completeness, a prominent, at first glace
+maybe a bit “dull” point on this Lie group is the identity matrix `e`.
+
+```{julia}
+SO3 = SpecialOrthogonalGroup(3)
+g = [cos(π/3) -sin(π/3) 0.0; sin(π/3) cos(π/3) 0.0; 0.0 0.0 1.0]* [cos(π/4) 0.0 -sin(π/4); 0.0 1.0 0.0; sin(π/4) 0.0 cos(π/4)]
```
mateuszbaran: Maybe it would be more clear to use Rotations.jl here? You could just do something like `g = RotZ(π/3) * RotY(π/4)`.

tutorials/getstarted.qmd
```diff
@@ -1,3 +1,312 @@
...
+e = Matrix{Float64}(I,3,3)
+g
+```
+The element `g` can be seen as a rotation by $π/3$ in the x-y-plane combines with a rotation by $π/4$ in the x-z plane.
+
+The tangent space $T_e𝒢$ plays a special role and is called the [`LieAlgebra`](@ref) $𝔤$.
+
+Similar to the ``[Riemannian exponential map](@extref `Base.exp-Tuple{AbstractManifold, Any, Any}`)``{=commonmark}
+The exponential function $\exp_𝒢: 𝔤 → 𝒢$ maps Lie algebra tangent vectors $X ∈ T_e𝒢$
+to a point on the Lie group. This is implemented in [`exp(G::LieGroup, X)`](@ref). It's inverse
```
mateuszbaran:
```suggestion
to a point on the Lie group. This is implemented in [`exp(G::LieGroup, X)`](@ref). Its inverse
```

tutorials/getstarted.qmd
```diff
@@ -1,3 +1,312 @@
...
+
+This allows to introduce an inner product on all tangent spaces, that smoothly varies with
+the base point, we obtain the left-invariant metric
+
+```math
+⟨X,Y⟩_g = ⟨D^*λ_g(e)[X],D^*λ_g(e)⟩_e,
+```
+
+since it is invariant if you use the differential of the left composition to identify tangent vectors.
+We can even “spare” the pullbacks in this definition, when spring tangent vectors as their representants in the Lie algebra.
```
mateuszbaran: spring?

tutorials/getstarted.qmd
```diff
@@ -1,3 +1,312 @@
...
+This is the default in `LieGroups.jl`.
```
mateuszbaran, line 158: Maybe it would be a good idea to add that this metric in general is not right-invariant, except for compact Lie groups and their direct products with `Translations`.

tutorials/getstarted.qmd
```diff
@@ -1,3 +1,312 @@
...
+With respect to this metric on the manifold the exponential and logarithmic maps are given by
+
+```math
+\exp_g(X) = g ∘ \exp_𝒢(X)
+\quad\text{with its inverse}\quad
+\log_g(h) = \log_𝒢(g^{-1}∘h)
+```
+
+With respect to these geodesics and the representation in the Lie algebra, parallel transport
+simplifies to the identity. To still access the Riemannian exponential map with recpect to the
```
mateuszbaran:
```suggestion
simplifies to the identity. To still access the Riemannian exponential map with respect to the
```

tutorials/getstarted.qmd
```diff
@@ -1,3 +1,312 @@
...
+metric (compatible to the Levi-cevita connection) on the underlying Riemannian manifold, use
```
mateuszbaran:
```suggestion
metric (compatible to the Levi-Civita connection) on the underlying Riemannian manifold, use
```

mateuszbaran (commented):
A few comments for the first part. I think this looks nice in general and only requires minor polishing.

tutorials/getstarted.qmd
kellertuer: Indeed. That is not nice. We could add a `\cdot`? But we could also leave it out, sure.

mateuszbaran: Either `\cdot` or nothing, depending on your preference :slightly_smiling_face:

tutorials/getstarted.qmd
kellertuer: After introducing the left invariant vector fields and identifying everything with the Lie algebra, I do agree – but I do that in the following

tutorials/getstarted.qmd
kellertuer: Formerly that is not necessary, since the rotation matrix does take care of that in cos/sin. 

I would prefer not writing a mod – but maybe adding that as a sentence that some might agree (have to think about a good phrasing).

tutorials/getstarted.qmd
kellertuer: Sure, fair enough. Makes the code more readable. Will do that.

tutorials/getstarted.qmd
kellertuer: ah autocorrect with too many typos. `representing`

tutorials/getstarted.qmd
kellertuer: Sure, maybe a bit of a technical note, since one has to say before also that right invariant can be build in a similar way. Will add that.

kellertuer:
>  I think this looks nice in general and only requires minor polishing.

Nice! Yes I always have a few typos here and there (and reading it in a week I would see them) – but good that you like the general idea. It took a bit of time coming up with a good level of detail and a good way through an example with code.

tutorials/getstarted.qmd
kellertuer: Ok let‘s do nothing for now then. That should also be clear.

tutorials/getstarted.qmd
mateuszbaran: Feel free to phrase it differently but I think it's important to make it more precise. We can either assume that $\beta=\alpha_1 + \alpha_2$ and conclude that the matrices multiply accordingly, or start from matrix multiplication (like you did) and conclude that we can only establish equality up to a multiple of $2\pi$.

kellertuer: I will think about it and agree, it should be more precise and clear.

tutorials/getstarted.qmd
kellertuer: Should we maybe add a reference for such a statement?

kellertuer:
Besides checking for a nice reference for left/right invariant vector fields yielding different metrics, this should be finished; thanks for the careful reading.

tutorials/getstarted.qmd
mateuszbaran: Sure. The most general statement of this kind that I know of is in
```julia
@article{LatifiToomanian:2013,
	title = {On the existence of bi-invariant {Finsler} metrics on {Lie} groups},
	volume = {7},
	issn = {2251-7456},
	doi = {10.1186/2251-7456-7-37},
	number = {1},
	journal = {Mathematical Sciences},
	author = {Latifi, Dariush and Toomanian, Megerdich},
	month = aug,
	year = {2013},
	pages = {37},
}
```
where they show that it's true not only for Riemannian metrics but also Finsler ones.

tutorials/getstarted.qmd
```diff
@@ -1,3 +1,324 @@
...
+X = log(SO3,g)
+is_point(LieAlgebra(SO3), X)
+```
+
+```{julia}
+exp(SO3,X)
+```
+
+The term “exponential” has at least three different meanings throughout Lie groups and Riemannian manifolds.
+To be precise, we call the just introduced one “exponential function”, since it often agrees with the [exponential](https://en.wikipedia.org/wiki/Exponential_function).
```
mateuszbaran:
```suggestion
To be precise, we call the just introduced one “exponential function”, since it often agrees with the [matrix exponential](https://en.wikipedia.org/wiki/Matrix_exponential).
```
It doesn't agree with the scalar exponential that often.

tutorials/getstarted.qmd
```diff
@@ -1,3 +1,324 @@
...
+from an element of $𝒢_1$. The group operation reads for $(g_1,g_2), (h_1,h_2) ∈ 𝒢$
+
+```math
+(g_1,g_2) ∘ (h_1,h_2) := (g_1 ⋆ h_1, σ_{h_1}(g_2) ⋄ h_2).
+```
+
+### Rigid body motions
+
+We obtain the [special Euclidean group](../groups/special_euclidean_group.md)
+$\mathrm{SE}(3) = \mathrm{SO}(3) ⋉ T(3)`` where the group action is the one discussed as an example before.
```
mateuszbaran:
```suggestion
$\mathrm{SE}(3) = \mathrm{SO}(3) ⋉ T(3)$ where the group action is the one discussed as an example before.
```

tutorials/getstarted.qmd
```diff
@@ -1,3 +1,324 @@
...
+gh1 = compose(SE3, g1, h1)
+```
+
+apply the Lie group exponential
+
+```{julia}
+Y = log(SE3, gh1)
+```
+
+and look at the manifold dimension, it is the sum of the single ones
```
mateuszbaran:
```suggestion
and look at the manifold dimension, it is the sum of dimensions of its components
```

tutorials/getstarted.qmd
```diff
@@ -1,3 +1,324 @@
...
+
+```{julia}
+manifold_dimension(SE3)
+```
+
+and its coordinates as
```
mateuszbaran:
```suggestion
The coordinates of `Y` are
```
I think this sentence is too long.

tutorials/getstarted.qmd
kellertuer: Nice, will add that then. That it is even more general is fine with me

mateuszbaran (approved):

PR #27, opened 2025-03-06
(An approach to) Linear representations

### LINKED ISSUE #6: New linear representation type
Many group actions are actually linear representations, which allows to define `apply_diff` automatically. I would suggest the following:

1. Create a new abstract type `AbstractLinearRepresentation`:
```julia
abstract type AbstractLinearRepresentation{AD <: ActionDirection} <: AbstractGroupAction{AD} end
```
2. Define `apply_diff` for this class of group actions:
```julia
Manifolds.apply_diff(A::AbstractLinearRepresentation, g, ::Any, X) = apply(A, g, X)
```
3. Make the existing linear group actions derive from `AbstractLinearRepresentation`
For instance:
```julia
struct TranslationAction{TAD<:ActionDirection,TM<:AbstractManifold,TRn<:TranslationGroup} <:
       AbstractLinearRepresentation{TAD}
    manifold::TM
    Rn::TRn
end
```
3. Clean up existing code
Most of the existing implementations of `apply_diff` follow this pattern (because they are linear representation), for instance:
- `apply_diff(::TranslationAction,...)`
- `apply_diff(::RotationActionOnVector,...)`
- `apply_diff(::RotationTranslationOnVector`,...)`

kellertuer:
This aims to resolve formerly issue 6 but I need more input here @olivierverdier 

* Literature for the docs so other can understand as well what this is for
* this “other” includes even me, since your issue explained _what_ you wanted to change in code but not _why_
* we do no longer have as many types as before, since there is now a generic `GroupOperationAction`
* this now acts as a wrapper for the action type the same way one indicates that a matrix is hermitian with `Hermitian(A)`
* 

Concrete things missing
* [ ] implement `default_group_action(L,M)` cases to return this wrapper for Lie group / manifold cases where this helps
  * Q1: which ones would that be? 
* Q2: Are there other functions besides the differential of applying the group action, that benefit from this as well?
* [ ] Documentation – Especially including Literature.
  * Q3: Please provide literature related to #6 
* [ ] Code coverage


olivierverdier:
First, here is why *you* may want to do this: it is to avoid repeating the same implementation of `apply_diff` many times. (See, e.g., the definitions of `apply_diff(::TranslationAction,...)`, `apply_diff(::RotationActionOnVector,...)`, `apply_diff(::RotationTranslationOnVector,...)`)

Why *I* wanted this is for my own group actions (which are also representation), so that I wouldn't have to reimplement `apply_diff`.

The documentation: it is simply a special case of a group action (a morphism from the group to a subset of bijections on a set), called [representation](https://en.wikipedia.org/wiki/Group_representation) (a morphism from the group to a subset of **linear** bijections on a vector space).

Is that somewhat clearer?

kellertuer:
Hi,
thanks, that helps a bit, but I am not sure we will have those three actions  again, we now have a  generic `GroupAction` see https://juliamanifolds.github.io/LieGroups.jl/stable/interface/actions/#LieGroups.GroupAction, which avoids the need to duplicate the structs you mentioned, so that we can distinguish them by type.

For now we have the `GroupOperationActionType`s (e.g. https://juliamanifolds.github.io/LieGroups.jl/stable/interface/actions/#LieGroups.LeftGroupOperationAction ) when L=M in the action above.

I think I would summarise your three examples as a `GroupOperationActionOnType`? (name pending better ideas ;)). A main idea of group actions having such a name type is to have them avoid the redundancy we had before.

So the duplication is basically gone with the “acting on” type that still needs a good name. “Wrapping” that in a linear type as I am currently sketching is probably still a nice thing, to get the `diff` for free.

The second sentence with “it is a special case of” I do not understand. Also that only speaks about representation, but the main part here is that it is linear (for the diff to be of that form)? I am not yet able to document that.
Is there more functions than `diff_apply` that get a nice form?

olivierverdier:
> The second sentence with “it is a special case of” I do not understand.

A group action is a morphism from the group to a subset of bijections on a **set**.

A representation is a morphism from the group to a subset of **linear** bijections on a **vector space**.

A vector space is a set, a linear bijection is a bijection, hence it is a special case.

kellertuer:
Yes, but currently none of our docs speak about representations, so that has to somehow be documented. I do not feel clever enough to phrase that documentation. But from your super short answer it seems I am maybe just too stupid for this.

Also it currently makes exactly one function available as a default implementation. Are there more functions that get nicer?
And sure I have to check which defaults one could best provide then.

So for now I have no clue how to continue here, but I will leave that open, maybe some time in the far future I have an idea. For now I do not.

olivierverdier:
It's ok, it's not super important. It was just an idea that i had to avoid code duplication. I can absolutely live with the current approach.

kellertuer:
Nice.

I do see the advantage in both reducing code duplication this might introduce and help users see, that their action has nice properties for a certain combination of L and M. Both code and property wise this might be nice to have – but only after I (or someone else who wants to take over) have a better understanding of it. This might also include that we first have to take representations more serious.

kellertuer:
I think it is maybe better to close this, the issue is still open, so if someone needs this we can start from there again (this PR is anyways only 4 lines for now).

PR #29, opened 2025-03-16
Port remaining groups (part 1/3): Quaternion Unitary

kellertuer:


test/groups/test_unitary_group.jl
```diff
@@ -1,43 +1,73 @@
...
+        g1 = quat(1.0)
+        g2 = quat(0.0, 1.0, 0.0, 0.0)
+        g3 = quat(0.0, 0.0, 1.0, 0.0)
+        properties = Dict(
+            :Name => "The quaternion unitary group",
+            :Points => [g1, g2, g3],
+            :Vectors => [quat(0.0)],
+            :Mutating => false,
+            :Rng => Random.MersenneTwister(),
+            :Functions => [rand, exp, log, show],
```
mateuszbaran: Could you also test `compose`, `inv`, `exp_fused` and the mutating case? Manifolds.jl currently seems to support all of these.

kellertuer: ^exp_fused` does not yet exist in LieGroups.jl – and by now I think it will take quite a time until it does, if _I_ should implement that, not this year, since I am loosing a bit of motivation here. I am growing a bit of the feeling that I just wasted half a year of programming on something no one will ever use with this package.

I can check for compose and inv but I am not sure how to approach the mutating case?

mateuszbaran: OK, then let's leave `exp_fused` out for now. The mutating case should work the same way as in Manifolds.jl -- the quaternion is put into `Ref` or one-element array with `fill`. I think at least one of these two approaches should work.

kellertuer: Sure, can try the zero-array/fill thing. At least add the tests today and if they fill try to find time to fix that throughout the week

kellertuer: Hm, well, compose and inv also do not work. Ok, Then this “ah this is easy” PR will also turn into a few weeks of work I fear.

I really wish at some point I am able to finish any PR in a finite amount of time again.

kellertuer:
for now I added all tests that still need to be passed and will check them next then.

kellertuer:
Hm, it seems there is a lot of stuff in Manifolds.jl missing especially for the 0-dim-arrays for the generic implementations here to work probperly

* in `UnitaryMatrices` for the Lie group
* in `SkewHermitianMatrices` for the Lie algebra.

For now due to methods missing there nearly all tests here for zero-dim quat-arrays fail.

kellertuer:
A definite plus point of the test suite is, it checks all scenarios automatically.
A definite negative point of the test point is, it hence spots all mean corner cases automatically as well ;)

kellertuer:
@mateuszbaran I hope this now covers all points you requested besides the `_fused` variant that first someone (else) has to introduce to this ackage.

mateuszbaran (approved):
LGTM. Tests are definitely useful for catching corner cases :slightly_smiling_face:

PR #30, opened 2025-03-16
Port remaining groups (part 2/3): Special Linear

kellertuer:
starts the last group we did not yet start with: Special Linear.

This is, however, missing a special manifold (extension) in `Manifolds.jl`, namely invertible matrices of determinant one.
We have the Unitary-det1-case already but have to add a similar matrix type to the invertible matrices as well, though this will mainly be an effect for the `check_point` (to additionally check for det 1) and `check_vector` to check for trace 0.


kellertuer:
This is now on par with the functionality we have for the analogue in `GroupManifold`s,
but one could check whether there exist formulae for `hat/vee` might be complicated with the trace 0 constraint

mateuszbaran:
Wouldn't `c=vec(X)[1:end-1]`  work as coordinates? Then `[c..., -sum(c)]` would be the reconstructed vector (up to reshaping).

kellertuer:
I am not so sure, how does that keep symmetry of the trace zero things?
Also for any set of coordinates the result would not be trace 0.

kellertuer:
Ah no we do not have symmetry, (sorry, started too early today) but your fix does not work, because instead of sum we have to set the entry to zero build the matrix and take -tr of that matrix.

kellertuer:
With trace your idea worked well for the real case. For the complex one I got a bit lost

1. in again type conversion / allocation problems
2. (yet again) in basis stuff of real and complex bases and which one is which – in one we would have to split the coifs in real/complex parts, the other not.

Not sure it is worth to add that for the complex case. Let me know.

For the real case I will document hat/vee the next few days then with the formulae.

mateuszbaran:
Right, I forgot about the trace part in the reconstruction. The complex variant is not very important I think so let's skip it for now.

kellertuer:
Nice! I will add the docs to `hat` and `vee` hopefully tomorrow then.

kellertuer:
There we are, the docs are completed and hence I hope this is finished. The last Lie group is also making good progress :)

kellertuer:
Hmpf updating docs makes randomly another Lie group fail. But this fits this week as well like last week. I seem to have a streak of things-never-work currently. Will check somewhen later back again then. no clue what caused this now.

src/groups/special_linear_group.jl
```diff
@@ -0,0 +1,111 @@
...
+    X[end] = 0
+    X[end] = -tr(X)
+    return X
+end
+
+_doc_hat_special_linear = """
+    X = hat(𝔤::LieAlgebra{ℝ,MatrixMultiplicationGroupOperation,<:SpecialLinearGroup}, c)
+    hat!(𝔤::LieAlgebra{ℝ,MatrixMultiplicationGroupOperation,<:SpecialLinearGroup}, X, c)
+
+Compute the hat map ``(⋅)^{\\wedge} : ℝ^{n^2-1} → 𝔤`` that turns a vector of coordinates `c``
```
mateuszbaran:
```suggestion
Compute the hat map ``(⋅)^{\\wedge} : ℝ^{n^2-1} → 𝔤`` that turns a vector of coordinates `c`
```

src/groups/special_linear_group.jl
```diff
@@ -0,0 +1,111 @@
...
+    n = ManifoldsBase.get_parameter(M.size)[1]
+    return print(io, "SpecialLinearGroup($n, $(𝔽); parameter=:field)")
+end
+
+_doc_vee_special_linear = """
+    c = vee(𝔤::LieAlgebra{ℝ,MatrixMultiplicationGroupOperation,<:SpecialLinearGroup}, X)
+    vee!(𝔤::LieAlgebra{ℝ,MatrixMultiplicationGroupOperation,<:SpecialLinearGroup}, c, X)
+
+Compute the vee map ``(⋅)^{\\vee}: $(_math(:𝔤)) →  ℝ^{n^2-1}`` that maps a tangent vector
+from the Lie algebra to a vector of coordinates `c``
```
mateuszbaran:
```suggestion
from the Lie algebra to a vector of coordinates `c`.
```

src/groups/special_linear_group.jl
```diff
@@ -0,0 +1,111 @@
...
+
+_doc_hat_special_linear = """
+    X = hat(𝔤::LieAlgebra{ℝ,MatrixMultiplicationGroupOperation,<:SpecialLinearGroup}, c)
+    hat!(𝔤::LieAlgebra{ℝ,MatrixMultiplicationGroupOperation,<:SpecialLinearGroup}, X, c)
+
+Compute the hat map ``(⋅)^{\\wedge} : ℝ^{n^2-1} → 𝔤`` that turns a vector of coordinates `c``
+into a tangent vector in the Lie algebra.
+
+The formula on the Lie algebra ``𝔤`` of the [`SpecialLinearGroup`](@ref)`(n)` is given by
+reshaping ``c ∈ ℝ^{n^2-1}`` into an ``n``-by``n``matrix ``X`` with the final entry `X[n,n]`
```
mateuszbaran:
```suggestion
reshaping ``c ∈ ℝ^{n^2-1}`` into an ``n``-by``n`` matrix ``X`` with the final entry `X[n,n]`
```

test/LieGroupsTestSuite.jl/LieGroupsTestSuite.jl
```diff
@@ -390,19 +390,19 @@ end
 Test  `diff_conjugate`
 """
 function test_diff_conjugate(
-    G::AbstractLieGroup, g, h, X; expected=missing, test_mutating::Bool=true
+    G::AbstractLieGroup, g, h, X; expected=missing, test_mutating::Bool=true, atol::Real=0
```
mateuszbaran:
```suggestion
    G::AbstractLieGroup, g, h, X; expected=missing, test_mutating::Bool=true, atol::Real=sqrt(eps(real(number_eltype(X)))),
```
0 is rarely a good `atol` to have.

mateuszbaran (approved):
Just a few minor things, everything else looks fine.

PR #32, opened 2025-04-22
Fix `SU` math representation

MasonProtter:
Seems to be a typo which manifests in the docs (ref https://juliamanifolds.github.io/LieGroups.jl/v0.1/groups/special_unitary_group/) which makes `SU(N)` display confusingly as `SO(N)`


kellertuer:
Well-spotted, thanks!

And yes, we introduced those short hands to only have one place where we fix the notation and here we made a copy-paste error, so thanks for directly fixing that :)

PR #34, opened 2025-04-29
Adapt to `default_basis` and pass onwards `get_coordinates`/`get_vector`

### LINKED ISSUE #33: missing dispatches for cov, e.g. get_coordinates_orthonormal
Hi, I think as LieGroups.jl builds out there might be a few extra dispatch routes not covered yet (I'm still getting to grips with new API). 
> OR I'm using the module namespaces wrong when mixing Manifolds and LieGroups during transition:

## Example Showing Error

```julia
using StaticArrays
using Statistics
import Manifolds as MF
import LieGroups as LGr


# copy paste of slightly long winded example from before Circle or Lie groups were stable in Manifolds
# idea is rotate and translate some 2D points before estimating their covariance through "on-manifold" dispatches
r_C = pi/3
ax_CC = [SA[5*randn();randn()] for _ in 1:100]
M = LGr.TranslationGroup(2)   # still works when using MF.TranslationGroup(2)
_R(α, s=exp(-α*im)) = real(s)*SA[1 0; 0 1] + imag(s)*SA[0 1; -1 0]
r_R_ax = _R(r_C)
# rotate coordinates
r_CC = map(ax_CC) do ax_C
  r_R_ax*ax_C + SA[10;-100]
end
r_CV = Statistics.cov(M, r_CC)  # error occurs on this line with cov
```

 ----

## Error output

```julia
ERROR: MethodError: no method matching get_coordinates_orthonormal!(::LieGroups.TranslationGroup{…}, ::Vector{…}, ::MVector{…}, ::MVector{…}, ::ManifoldsBase.RealNumbers)
The function `get_coordinates_orthonormal!` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  get_coordinates_orthonormal!(::Manifolds.PositiveNumbers, ::Any, ::Any, ::Any, ::ManifoldsBase.RealNumbers)
   @ Manifolds ~/.julia/packages/Manifolds/nD417/src/manifolds/PositiveNumbers.jl:171
  get_coordinates_orthonormal!(::Circle{ℝ}, ::Any, ::Any, ::Any, ::ManifoldsBase.RealNumbers)
   @ Manifolds ~/.julia/packages/Manifolds/nD417/src/manifolds/Circle.jl:189
  get_coordinates_orthonormal!(::Circle{ℂ}, ::Any, ::Any, ::Any, ::Union{ManifoldsBase.ComplexNumbers, ManifoldsBase.RealNumbers})
   @ Manifolds ~/.julia/packages/Manifolds/nD417/src/manifolds/Circle.jl:211
  ...

Stacktrace:
  [1] get_coordinates_orthonormal(M::LieGroups.TranslationGroup{…}, p::MVector{…}, X::MVector{…}, N::ManifoldsBase.RealNumbers)
    @ ManifoldsBase ~/.julia/packages/ManifoldsBase/AinsU/src/bases.jl:572
  [2] _get_coordinates(M::LieGroups.TranslationGroup{…}, p::MVector{…}, X::MVector{…}, B::ManifoldsBase.DefaultOrthonormalBasis{…})
    @ ManifoldsBase ~/.julia/packages/ManifoldsBase/AinsU/src/bases.jl:566
  [3] get_coordinates(M::LieGroups.TranslationGroup{…}, p::MVector{…}, X::MVector{…}, B::ManifoldsBase.DefaultOrthonormalBasis{…})
    @ ManifoldsBase ~/.julia/packages/ManifoldsBase/AinsU/src/bases.jl:541
  [4] (::Manifolds.var"#97#98"{…})(X::MVector{…})
    @ Manifolds ~/.julia/packages/Manifolds/nD417/src/statistics.jl:53
  [5] iterate
    @ ./generator.jl:48 [inlined]
  [6] _collect
    @ ./array.jl:810 [inlined]
  [7] collect_similar
    @ ./array.jl:732 [inlined]
  [8] map
    @ ./abstractarray.jl:3361 [inlined]
  [9] cov(M::LieGroups.TranslationGroup{…}, x::Vector{…}; basis::ManifoldsBase.DefaultOrthonormalBasis{…}, tangent_space_covariance_estimator::StatsBase.SimpleCovariance, mean_estimation_method::ManifoldsBase.GradientDescentEstimation, inverse_retraction_method::ManifoldsBase.LogarithmicInverseRetraction)
    @ Manifolds ~/.julia/packages/Manifolds/nD417/src/statistics.jl:53
 [10] cov(M::LieGroups.TranslationGroup{ℝ, ManifoldsBase.TypeParameter{Tuple{2}}}, x::Vector{SVector{2, Float64}})
    @ Manifolds ~/.julia/packages/Manifolds/nD417/src/statistics.jl:35
 [11] top-level scope
    @ REPL[32]:1
```

cc @kellertuer  
- just [an xref to mention on 21](https://github.com/JuliaManifolds/LieGroups.jl/issues/21#issuecomment-2834015014)
- example origin: https://github.com/JuliaRobotics/ApproxManifoldProducts.jl/blob/06af24580c60e46892d162f9176a591bc516b29c/test/manellic/testManifoldTreeConstr.jl#L37

kellertuer:
This hopefully fixes #33, at least on my machine the example runs fine.

_edit:_ requires https://github.com/JuliaManifolds/ManifoldsBase.jl/pull/229


mateuszbaran (approved):
LGTM

kellertuer:
Ah I did not yet run the tests, there are some strange ambiguities with some RiezRepresenterTangentVector stuff, I think that is from your design?

src/interface.jl
```diff
@@ -430,6 +433,21 @@ end
 @doc "$(_doc_exponential)"
 ManifoldsBase.exp!(G::AbstractLieGroup, ::Any, ::Any)
 
+function ManifoldsBase.get_coordinates(G::AbstractLieGroup, g, X, B::AbstractBasis)
```
```suggestion
function ManifoldsBase.get_coordinates(G::AbstractLieGroup, g, X, B::AbstractBasis{<:Any,TangentSpaceType})
```
This restricts the method to based of tangent spaces (rather than arbitrary bases of vector spaces) and should fix the ambiguity because the ambiguous methods are for the cotangent space.

mateuszbaran:
Yes, I've suggested a fix.

src/interface.jl
kellertuer: Ah, I See. That detail of basis I did not have in my active memory.

kellertuer:
That resolved one of the 4 ambiguities but the other 3 will work the same way for the other 3 functions. I can do that tomorrow (currently not close to a VS Code).

kellertuer:
I hope all is fine now with this; @dehann do you want to check this out before we release this? Your posted example does work with this :)

I also forgot to activate zenodo for the last release but did it by now. Will edit the Readme accordingly after the next release (and a first zenodo) is available.

dehann:
Hi @kellertuer , thanks for the update.  I tried and `cov` function is working with the tests I have! :+1: 

I did come across another missing dispatch though, on `identity_element`.  Looks like `Manifolds.identity_element` won't work with `LieGroups.TranslationGroup` -- am I using it wrong?  My sense was dispatch would resolve the best available function in either Manifolds or LieGroups?

```julia
ERROR: MethodError: no method matching identity_element(::LieGroups.TranslationGroup{ℝ, ManifoldsBase.TypeParameter{Tuple{2}}}, ::SVector{2, Float64})
The function `identity_element` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  identity_element(::Manifolds.RealCircleGroup, ::Any)
   @ Manifolds ~/.julia/packages/Manifolds/nD417/src/groups/circle_group.jl:220
  identity_element(::LieGroups.AbstractLieGroup)
   @ LieGroups ~/.julia/packages/LieGroups/Ji7lz/src/interface.jl:475
  identity_element(::LieGroups.LieGroup{𝔽, LieGroups.AdditionGroupOperation, M} where M<:AbstractManifold{𝔽}, ::Type{Array{T, 0}}) where {𝔽, T<:Number}
   @ LieGroups ~/.julia/packages/LieGroups/Ji7lz/src/group_operations/addition_operation.jl:146
  ...
```

kellertuer:
Thanks for checking. 

Interesting! Do you maybe have a small code snippet to check with? Because here the following lines should have been enough to get you covered

https://github.com/JuliaManifolds/LieGroups.jl/blob/56650add94bb4dfca7dd4da136b60a65c02d4432/src/group_operations/addition_operation.jl#L141-L153

So if these are not hit, I do need an example to check how to best generalise those lines. Ah! I think I see what is wrong (maybe): `Identity_element` would work without a second argument and use the default representation of points then,
if you provide a second argument you have to provide the _type_. and then get new memory (containing the identity) of that type. cf. 

If you have a point `g` on the Lie group `G`  already and would like to “fill” that point (your `SArray` maybe?) with the identity, use `identity_element!(G,g)`.
Both cases are described in https://juliamanifolds.github.io/LieGroups.jl/stable/interface/group/#Manifolds.identity_element-Tuple{AbstractLieGroup}

kellertuer:
If this is different from the previous behaviour (the old docs at https://juliamanifolds.github.io/Manifolds.jl/stable/manifolds/group/#Manifolds.identity_element-Tuple{AbstractDecoratorManifold,%20Any} are a bit imprecise for me right now) we should add that as a comment to the transition tutorial at https://juliamanifolds.github.io/LieGroups.jl/stable/tutorials/transition/#Further-notable-changes

dehann:
Hi, just catching up here -- thanks for feedback.  Let me have a look and if not an obvious issue this side I'll get a code snippet...

dehann:
> docs bit ambiguous

Yeah, I'd agree the `p` vs `T` on "point type".  From the docs: "where p indicates the type to represent the identity".

# Reproduce (notice return types change)

> example

## Via LieGroups.jl

```julia
using LieGroups, StaticArrays

M = TranslationGroup(2)
e = SVector(0,0.)

# this works but return type changes
identity_element(M,typeof(e))
# 2-element Vector{Float64}:
# 0.0, 0.0

# this fails (error stack below)
identity_element(M,e)
```

## Via Manifolds.jl

```julia
using Manifolds, StaticArrays

M = TranslationGroup(2)
e = SVector(0,0.)

# this works but has MVector return type
identity_element(M,typeof(e))
# 2-element MVector{2, Float64} with indices SOneTo(2):
#  0.0 0.0

# this works on Manifolds but has MVector return type
identity_element(M,e)
# 2-element MVector{2, Float64} with indices SOneTo(2):
#  0.0 0.0
```

----

## Error on point type using LieGroups
```julia
# from LieGroups.jl example
identity_element(M,e)

ERROR: MethodError: no method matching identity_element(::TranslationGroup{ℝ, ManifoldsBase.TypeParameter{Tuple{2}}}, ::SVector{2, Float64})
The function `identity_element` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  identity_element(::Manifolds.RealCircleGroup, ::Any)
   @ Manifolds ~/.julia/packages/Manifolds/nD417/src/groups/circle_group.jl:220
  identity_element(::LieGroup{𝔽, AdditionGroupOperation, M} where M<:ManifoldsBase.AbstractManifold{𝔽}, ::Type{Array{T, 0}}) where {𝔽, T<:Number}
   @ LieGroups ~/.julia/packages/LieGroups/Ji7lz/src/group_operations/addition_operation.jl:146
  identity_element(::LieGroup{𝔽, AdditionGroupOperation, M} where M<:ManifoldsBase.AbstractManifold{𝔽}, ::Type{T}) where {𝔽, T<:Union{Number, AbstractArray{0, <:Number}}}
   @ LieGroups ~/.julia/packages/LieGroups/Ji7lz/src/group_operations/addition_operation.jl:141

```

Edit: mentioning nitpick: first of closest candidates takes for second argument `::Any`.  Might be confusing down the road.

kellertuer:
If that is ok, we would like to discontinue the second signature that worked still in Manifolds.jl and only keep the one with the `type`, since that is more consistent with the rest of the interfaces, `zero_vector` for example.

I will try to find out where we “loose” your type and fix that in LieGroups.jl, that I would consider definitely a bug.

kellertuer:
Also note that all functionality of Lie groups per se, like `compose` or `identity_element` are now deprecated in Manifolds.jl and will be removed on the next breaking release – though we will wait with that release a git until we are sure LieGroups.jl is relatively stable.
After that release the note in our Readme here will also be removed again, then they do not define a few things double :)

I saw that in https://github.com/JuliaRobotics/ApproxManifoldProducts.jl/pull/313/files quite a few of the deprecated functions might still be used.

kellertuer:
> Edit: mentioning nitpick: first of closest candidates takes for second argument ::Any. Might be confusing down the road.

yes, that is still from the `Manifolds.jl` variants of the function that are deprecated, since there it was vague and often both the point and its type were accepted. I hope in LieGroups.jl it is unified to one thing: the type.

kellertuer:
@dehann I added a note to the `transition.md` page about the second argument of `identity_element`, i.e that your second call is now expected to fail on `LieGroups`, for the accidental change of type, Mateusz was kind enough to fix that in #35.

_edit:_ That fix is merged into main and already merged back here as well, so from your example above only the case we no longer plan to support (passing an actual point instead if its type) fails. That this is expected to fail is now documented on the transition page.

dehann:
> If that is ok, we would like to discontinue the second signature that worked still in Manifolds.jl and only keep the one with the type, since that is more consistent with the rest of the interfaces, zero_vector for example.
> for the accidental change of type, Mateusz was kind enough to fix that in https://github.com/JuliaManifolds/LieGroups.jl/pull/35.

Fine by me, and thanks again!

----

# Aside

> all functionality of Lie groups per se, like compose or identity_element are now deprecated in Manifolds.jl and will be removed on the next breaking release
> quite a few of the deprecated functions might still be used.

Thanks also for heads up.  We still have to go do the work to stop using the `HybridTangentRepresentation` and switch to `LeftInvariantRepresentation` (Mateusz mentioned the other day).  We had some trouble during our first attempts with `exp` and `log` and have to go back to resolve that proper.  I'm still tripping over general robotics usage vs. non-robotics usage and need to spend more time with the conventions to get a better grip on whats happening (I'm reading the raw Manifolds.jl src towards a more concise understanding).  But ... that's a separate story from this PR thread.

kellertuer:
I have never understood the hybrid form too much, because I was not involved in that, but it is probably really best to do that in a new issue and discuss what to do best.

One hybrid thing I could image is to do a `ProducManifold` of a Lie group and a manifold? but then your second part of that is not a Lie group.
Well, I would need more details on that I think – in the new issue. yeah, please open that when you feel stuck – I am also not too far into robotics-lingua here (yet?), so probably good for both of us to learn something :)

But then I'll first merge this and register a 0.1.1

mateuszbaran:
Yes, I imagine switching from the hybrid to the left invariant representation might be a bit involved but in the long run it should simplify the logic and code. In the short term you may just have functions to convert back and forth but in the long term I'd suggest working in the left-invariant representation as much as possible.

PR #37, opened 2025-05-08
SE(n) improvements

mateuszbaran:
`point_type_to_vector_type` probably should be in ManifoldsBase.jl? I was surprised such function doesn't exist yet.


src/interface.jl
```diff
@@ -445,10 +445,20 @@ function ManifoldsBase.get_coordinates!(
     return c
 end
 
+"""
+    point_type_to_vector_type(M::AbstractManifold, TP)
+
+Change point type `TP` that represents point on manifold `M` to matching type for
+representing tangent vectors.
+"""
+point_type_to_vector_type(::AbstractManifold, TP) = TP
```
kellertuer: I think until now we did that implicitly, e.g. with `zero_vector(M, p)` we used the type of `p` to generate the right tangent vector type `T`, see for example https://github.com/JuliaManifolds/Manifolds.jl/blob/d55acb407b44298ed077a4edb2d9c875474538fb/src/manifolds/FixedRankMatrices.jl#L838-L853

kellertuer: For the name, since the point type is a parameter, we could maybe to /even both)

`tangent_vector_type(M, point_type::Type)`
and its opposite direcection
`point_type(M, vector_type::Type)`
in naming?

The functionality sounds very reasonable, for sure :)

kellertuer: (or one could even leave our the `_type`, which leaves the opportunity, that a concrete `tangent_vector(M, p)` could return an allocated `X`.

Ah but then the point thingy should maybe be called `manifold_point(M,tp)`?

mateuszbaran: Yes, it looks like we didn't need it because we didn't pass around those types. I can add `tangent_vector_type(M, point_type::Type)` to ManfoldsBase.jl. The other direction is probably not necessary for now.

kellertuer: sure `tangent_vector_type` should for now be enough and sure it could be part of `ManifoldsBase`, though then we should be careful to provide that for existing pairs maybe.

mateuszbaran: Yes, it will have to be populated with existing pairs.

kellertuer:
LGTM, but we could add the note that the default `SE(3)` is now RxT not TxR and one should use `TranslationGroup(3)⋊SpecialOrthogonal(3)` at the bottom of the `transition.md` list under notable changes.

mateuszbaran:
Sure, can you add that to the PR where you think it fits?

kellertuer (approved):
Sure, I added a seventh point, please read it; I hope it is detailed enough – with that this is ready to go for me.

PR #42, opened 2025-05-29
Towards consistent return types

### LINKED ISSUE #39: Inconsistent return type for rand on SE2
rand returns an ArrayPartition where as identity_element returns a homogenous matrix. 
```julia
julia> G = SpecialEuclideanGroup(2; variant=:right)
SpecialEuclideanGroup(2; variant=:right)

julia> rand(G)
([0.44690164157224876, -0.6995448763784617], [0.4075161499023314 0.913198000199727; -0.913198000199727 0.4075161499023314])

julia> identity_element(G)
3×3 Matrix{Float64}:
 1.0  0.0  0.0
 0.0  1.0  0.0
 0.0  0.0  1.0
```

kellertuer:
Once finished, should fix #39.

For now this PR does achieve

```
julia> using LieGroups

julia> G = SpecialEuclideanGroup(2; variant=:right)
SpecialEuclideanGroup(2; variant=:right)

julia> identity_element(G)
3×3 Matrix{Float64}:
 1.0  0.0  0.0
 0.0  1.0  0.0
 0.0  0.0  1.0

julia> rand(G)
3×3 Matrix{Float64}:
  0.313702  0.949522   0.309771
 -0.949522  0.313702  -0.994399
  0.0       0.0        1.0
```

but it is more tricky, since we had a variant of `identity_element(G,T)` that dispatched just to drop `T`. That led to `T=SpecialEuclideanProductPoint` returning the matrix when `RecursiveArrayTools` are not loaded – as well as when it was loaded, since the extension only does that correctly if one also provides the internal `ArrayPartition` _including_ its parameters. 

That I could not yet resolve and would need some help with or a few weeks to ponder about.
It seems the default “drop `T`” is maybe not so good of a default, but dropping that for now lets about 400 tests fail – so also there I am a bit lost what to do.

So for now this is a very rough draft of what might fix the above mentioned issue since I have no idea how to continue, but I tried for most of the afternoon.

### DIFF
--- a/src/group_operations/addition_operation.jl
+++ b/src/group_operations/addition_operation.jl
@@ -139,6 +139,11 @@ which for the [`AdditionGroupOperation`](@ref) is the zero element or array.
 identity_element(::LieGroup{𝔽,AdditionGroupOperation}) where {𝔽}
 
 function identity_element(
+    G::LieGroup{𝔽,AdditionGroupOperation}, ::Type{T}
+) where {𝔽,T<:AbstractArray}
+    return zeros(representation_size(G.manifold))
+end
+function identity_element(
     ::LieGroup{𝔽,AdditionGroupOperation}, ::Type{T}
 ) where {𝔽,T<:Union{Number,AbstractArray{<:Number,0}}}
     return zero(T)
--- a/src/group_operations/multiplication_operation.jl
+++ b/src/group_operations/multiplication_operation.jl
@@ -218,10 +218,10 @@ which for an [`AbstractMultiplicationGroupOperation`](@ref) is the one-element o
 identity_element(::LieGroup{𝔽,<:AbstractMultiplicationGroupOperation}) where {𝔽}
 
 function identity_element(
-    G::LieGroup{𝔽,<:AbstractMultiplicationGroupOperation}, ::Type{T}
-) where {𝔽,T<:AbstractArray}
+    G::LieGroup{𝔽,<:AbstractMultiplicationGroupOperation}, ::Type{<:AbstractArray{T,2}}
+) where {𝔽,T}
     (N, M) = representation_size(G.manifold)
-    return Matrix{Float64}(I(N))
+    return Matrix{T}(I(N))
 end
 
 @doc "$(_doc_identity_element_mult)"
--- a/src/group_operations/multiplication_operation.jl
+++ b/src/group_operations/multiplication_operation.jl
@@ -217,6 +217,13 @@ which for an [`AbstractMultiplicationGroupOperation`](@ref) is the one-element o
 @doc "$(_doc_identity_element_mult)"
 identity_element(::LieGroup{𝔽,<:AbstractMultiplicationGroupOperation}) where {𝔽}
 
+function identity_element(
+    G::LieGroup{𝔽,<:AbstractMultiplicationGroupOperation}, ::Type{T}
+) where {𝔽,T<:AbstractArray}
+    (N, M) = representation_size(G.manifold)
+    return Matrix{Float64}(I(N))
+end
+
 @doc "$(_doc_identity_element_mult)"
 identity_element!(::LieGroup{𝔽,<:MatrixMultiplicationGroupOperation}, e) where {𝔽}
 function identity_element!(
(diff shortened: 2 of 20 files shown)


src/group_operations/addition_operation.jl
```diff
@@ -138,6 +138,11 @@ which for the [`AdditionGroupOperation`](@ref) is the zero element or array.
 @doc "$(_doc_identity_element_add)"
 identity_element(::LieGroup{𝔽,AdditionGroupOperation}) where {𝔽}
 
+function identity_element(
```
kellertuer, line 141: I think this would only be consistent if `T` is an `<:AbstractArray` and er can pass its eltype or even its type to zeros?

src/group_operations/multiplication_operation.jl
```diff
@@ -217,6 +217,13 @@ which for an [`AbstractMultiplicationGroupOperation`](@ref) is the one-element o
 @doc "$(_doc_identity_element_mult)"
 identity_element(::LieGroup{𝔽,<:AbstractMultiplicationGroupOperation}) where {𝔽}
 
+function identity_element(
```
kellertuer, line 220: same here. I would always expect `identity(G, T)`that the result is either of type `T` if that is possible, or an error otherwise (if `T` is a string or a type that does not fit).

src/group_operations/addition_operation.jl
mateuszbaran: Right, it would be better to pass the eltype. Note, however, that returning `Array` instead of a different array type does make some sense. For example if someone passed a `SubArray` here, it wouldn't be reasonable to try to return a `SubArray`.

kellertuer: We have to pass the main type as well to know whether we should e.g. create the affine matrix or the ArrayPartition variant, but sure the eltype might also be a good idea to pass. Not fully sure yet, how. And interesting that we did not yet stumble into that earlier with random point (e.g. on Hyperbolic).

mateuszbaran: `ArrayPartition` is special-cased anyway so it should work fine.

kellertuer: What should work fine ;) ? Do you propose to switch the `T` to an `eltype`? Then we can not easily distinguish different point representations (because they can have the same eltype)

mateuszbaran: The thing that should work fine is defaulting to `Array` with the right `eltype` for general `AbstractArray`s. `ArrayPartition` and our special point types will work because they will be special-cased.

mateuszbaran: Note that for example `similar` in base Julia also defaults to returning `Array`: https://github.com/JuliaLang/julia/blob/75946ce31ccc4c32979e6a0fec5daabe20f8a58a/base/abstractarray.jl#L833 .

kellertuer: Sure, mimicking that sounds fine to be, because it resolves the SubArray problem you mentioned and staying consistent with what similar does is a good idea :)

kellertuer:
* I checked the two discussions and agree that the current ones should be fine since they are consistent with `similar`.
* I also made sure the tests are running again and checked the docs.
* I checked the extension that the ArrayPartition types work well
* I made all tests work again with the new signature.

In between I felt it was necessary to introduce `point_type(T)` that we could maybe move to `ManifoldsBase`, to be precise
`zero_vector(𝔤, T)` on a Lie algebra has the “problem” that `𝔤` only stores `Identity` (cool for memory) but hence does not know the type `P` of a point. However, this is now passed to `identity_element(base_manifold(𝔤), P)`, so pro forma we need `point_type(G, T)`to get this `P`.
In practice it is not yet really used after all, but this seems “more correct”. 
This took a bit of time in the validation case, but I feel that this special case made also the other code nicer.

This is slightly breaking, since we had 2 Lie groups that used `identity_element(G, g::P)` to determine a point type instead of `identity_element(G, P)`.

I would consider this finished now.

kellertuer:
This should also fix all of #39, since that depends on `identity_element(G,P)` and `zero_vector(𝔤, T)` and both are not respecting their types `P` and `T` thoroughly.

mateuszbaran (approved):
I've expanded NEWS a bit and added a couple of tests. I think it's good now.

kellertuer:
Oh I did not necessarily mean to change a default, but yeah unifying that seems reasonable. Since this is not urgent, I'll just merge it and release it with the other PR once that is done.

PR #47, opened 2025-06-22
Use retractions, inverse retractions or vector transports on the base manifold

### LINKED ISSUE #43: Using retractions on the underlying manifold
A nice idea that could be useful is to implement a way to use any retraction (or also the exponential map) on the underlying Riemannian manifold as a retraction on the LIeGroup.

The idea would be to define a `BaseManifoldRetraction(retraction::AbstractRetractionMethod)` that would perform the following steps when calling
`retract(G, g, X, base_manifold_retraction)`:

1. push forward `X` from the Lie algebra to T_gG. With the current default of left-invariant vector fields it boils down to calling https://juliamanifolds.github.io/LieGroups.jl/stable/interface/group/#LieGroups.diff_left_compose-Tuple{AbstractLieGroup,%20Any,%20Any,%20Any} – call the result `Y`
2. (eventually) materialise `g` if it is the `Identity` – say in `g_`.
3. call `h = retract(base_manifold(G), g_, Y, base_manifold_retraction.retraction)`
 4. return `h`.

This would make it easy to for example “decouple” SE(3) into SO(3) x Euclidean(3) if an `ExponentialRetraction` thereon makes sense.
cc @Affie 

And just when I wrote the last sentence, maybe the push forward is not 100% easy to do since for the `RotationMatrices` I think the tangent space is again using the Lie algebra, but we will find a way I am sure :)

kellertuer:
I sketched the first idea to resolve #43.
This is WIP, since my first idea was that it might work without a specific “wrapper type” for these, but it creates too many ambiguities on Layer 2, so the wrapper is necessary and we have to move it to Layer 3; will work on that next then.
So this is a draft for now.

This is not really related to the followup discussions in the linked issue but resolves the original request.

# 📋
* [x] documentation of the base manifold pass functions
* [x] test coverage
* [x] news entry.
* [x] update the docs and adapt the last fixed I did (especially to move to the correct `diff_group_apply`)
* [x] refactor the code to have a single function to extract (a) the two Lie groups (b) the underlying product manifold (c) the group action, call it `product_lie_group_elements`.
* [x] provide converts between ArrayPartition and ArrayMatrix.


mateuszbaran:
I think this looks roughly OK. The wrappers are probably necessary to not to avoid ambiguity hell. I think it would be good to add a paragraph to docs about the interface that would explain this feature.

kellertuer:
Sure docs would come next. I had an hour to spare to check what would need to be done; moving it now from Layer 2 to 3 is the next step, but sure then docs. will work on that next.

kellertuer:
With Layer 3, tests run fine, now I just have to check how to best test it, or to be precise, what to write for the test suite.

kellertuer:
I noticed a problem with this one: If the manifold also uses a Lie-algebra-like tangent vector representation, like Rotations or the SPDs, then this would do the wrong thing. I am not yet sure how to resolve that.

mateuszbaran:
We probably need a new function that tells if a Lie group modifies the representation of tangent vectors or not.

kellertuer:
Looks good, just that the now uncovered code was previously covered by cases that were not named tuples but single functions in the F-type of the objective. Those still have to be checked / adapted.

kellertuer:
yes, I till think about it, mainly a change from Lie algebra to TpM that (similar to the last step in project) changes representation.

kellertuer:
We now have two new functions `push_forward_tangent` and `pull_bacvk_tangent` that do the two things we need, I also already documented them.
I also already set the rotations case to the idenity;

I updated the initial post what is still missing, but that is mainly a bit of documentation and testing.

Before that feedback on the naming :)

kellertuer:
Besides a good idea which LieGroup / retraction (inverse retraction, vector transport) to test this with for code coverage, this Pr is finished.

mateuszbaran:
I think testing on `SpecialEuclidean` with `ProductRetraction`, `InverseProductRetraction` and `ProductVectorTransport` would make sense for example.

kellertuer:
Good point. I will remove the ready tag now, since the new methods do not yet work on product or power manifolds, so this is far from ready.

kellertuer:
I even first have to think what the right order for that case is and what to push where and such. So thanks for the note, this PR will then only continue after I had those ideas figured out.

kellertuer:
So, I completed the test, but it still errors, since I have not yet found how to provide

`diff_left_compose` either on SE(n) specifically or even on semidirect product groups. Do you maybe have an idea for a source where one could find something like that @mateuszbaran ?

mateuszbaran:
For generic semidirect products I'm not sure how to do that but for SE(n) you should be able to derive `diff_left_compose` from the embedding in GL(n+1) (then the generic `diff_left_compose!` on `AbstractMultiplicationGroupOperation` should work fine).

Affie:
Is it not [translate_diff]( https://github.com/JuliaManifolds/Manifolds.jl/blob/453e6367c1eecc8b309c05234b88ca332fc4e943/src/groups/group.jl#L992) from Manifolds.jl? Either the adjoint or the vector unchanged?
Sorry if I misunderstood the issue, don't want to be in the way here.

mateuszbaran:
It is roughly the same thing but LieGroups.jl has different internal dispatching. Though right now I'm not entirely sure why LieGroups.jl doesn't seem to default to computing `diff_left_compose` through `adjoint_action` like Manifolds.jl does for the `LeftInvariantRepresentation`.

kellertuer:
I had thought/hoped I had programmed that, but maybe not for semidirect products. For me this area is the most complicated, most technical part with the adjoint and such. Maybe just actions reached the same level of technicallity, where I am probably not clever enough to understand this.

So I would have to check about that fallback, that I thought I had implemented...

_edit:_ so it is this function
https://github.com/JuliaManifolds/Manifolds.jl/blob/453e6367c1eecc8b309c05234b88ca332fc4e943/src/groups/semidirect_product_group.jl#L149-L183

I missed? Might check to transfer that today in the evening then.

mateuszbaran:
No, that's a function for the hybrid representation (note `SemidirectProductGroupHVR`). In Manifolds.jl the dispatch goes through, I think, `adjoint_action!(G::SpecialEuclidean, Y, p, Xₑ, ::LeftAction)` and probably
```julia
translate_diff!(
    ::TraitList{<:IsGroupManifold{<:AbstractGroupOperation,LeftInvariantRepresentation}},
    G::AbstractDecoratorManifold,
    Y,
    p,
    ::Any,
    X,
    ::LeftBackwardAction,
)
```
edit: I mean the dispatch for the left-invariant representation.

mateuszbaran:
So I think we need `diff_left_compose` implemented by default through `adjoint_action!` and `adjoint_action!` implemented for `SpecialEuclideanGroup`. Both are present in Manifolds.jl.

kellertuer:
Thanks for the clarification, yes. Probably something like that. Just that I first have to carefully dive into the jungle of functions over in Manifolds.jl – here in LieGroups, I hope I at least have a bit of an overview what is where and what for – in Manifolds.jl there were also a bit too many of these (see my wrong choice above).

kellertuer:
Hm, so, I am stuck again trying to match the old code against my best of knowledge of what needs to be done.
I can add more documentation and formulae, but for now I do not find a good formula that would help me here.

In the old code there is `adjoint_action` – and as far as I see it, we should use this one?

https://github.com/JuliaManifolds/Manifolds.jl/blob/9cc37787aaf67130dfd29ceeda757749e2068510/src/groups/special_euclidean.jl#L859-L871

But that internally also uses `adjoint_action`? It is not documented what it should do (no formula?) There is one above that explains a bit in words what it does but that is the Hybrid variant we no longer do.

Should that then be used to implement `diff_lect_compose` which was previously maybe 

https://github.com/JuliaManifolds/Manifolds.jl/blob/9cc37787aaf67130dfd29ceeda757749e2068510/src/groups/special_euclidean.jl#L841-L857

But since that is also undocumented and we do switch a representation in between I think (?) I am a bit lost and do not how to continue here.
I simply do not get it and feel tremendously stupid again, because I am now code-acheologizing for an hour again.

@mateuszbaran can you take a look? Sorry to bother you again. 

_edit:_ ah it is probably also not the diff I linked because that is BackwardRightWithershinsDownwards ( ;) ) which is not our default left.
Maybe checking literature is then a good next step to first document a formula. Where were these implementations / formulae from?

src/groups/semidirect_product_group.jl
```diff
@@ -272,6 +272,130 @@ function _compose!(
...
+    )
+    return Y
+end
+
+_doc_RSDP_diff_left_compose = """
+    diff_left_compose(
+        SDPG::LieGroup{𝔽,RightSemidirectProductGroupOperation,<:ProductManifold}, X, g, h
+    ) where {𝔽}
+    diff_left_compose!(
+        SDPG::LieGroup{𝔽,RightSemidirectProductGroupOperation,<:ProductManifold}, X, g, h
```
mateuszbaran:
```suggestion
        SDPG::LieGroup{𝔽,RightSemidirectProductGroupOperation,<:ProductManifold}, g, h, X
    ) where {𝔽}
    diff_left_compose!(
        SDPG::LieGroup{𝔽,RightSemidirectProductGroupOperation,<:ProductManifold}, Y, g, h, X
```
That's how it is defined in the code below.

src/groups/semidirect_product_group.jl
```diff
@@ -272,6 +272,130 @@ function _compose!(
     return k
 end
 
+_doc_LSDP_diff_left_compose = """
+    diff_left_compose(
+        SDPG::LieGroup{𝔽,LeftSemidirectProductGroupOperation,<:ProductManifold}, X, g, h
+    ) where {𝔽}
+    diff_left_compose!(
+        SDPG::LieGroup{𝔽,LeftSemidirectProductGroupOperation,<:ProductManifold}, X, g, h
+    ) where {𝔽}
```
mateuszbaran:
```suggestion
    diff_left_compose(
        SDPG::LieGroup{𝔽,LeftSemidirectProductGroupOperation,<:ProductManifold}, g, h, X
    ) where {𝔽}
    diff_left_compose!(
        SDPG::LieGroup{𝔽,LeftSemidirectProductGroupOperation,<:ProductManifold}, Y, g, h, X
    ) where {𝔽}
```

kellertuer: Ah yes, thanks :) I wrote the code of the docs when I was still a bit confused. I hope the math is better.

src/groups/special_orthogonal_group.jl
```diff
@@ -103,6 +103,9 @@ ManifoldsBase.log(::SpecialOrthogonalGroup{ManifoldsBase.TypeParameter{Tuple{4}}
 @doc "$(_doc_log_O4_id)"
 ManifoldsBase.log!(::SpecialOrthogonalGroup{ManifoldsBase.TypeParameter{Tuple{4}}}, X, g)
 
+pull_back_tangent!(::SpecialOrthogonalGroup, Y, g, X; kwargs...) = (Y .= X)
+push_forward_tangent!(::SpecialOrthogonalGroup, Y, g, X; kwargs...) = (Y .= X)
```
mateuszbaran:
```suggestion
pull_back_tangent!(::SpecialOrthogonalGroup, Y, g, X; kwargs...) = copyto!(Y, X)
push_forward_tangent!(::SpecialOrthogonalGroup, Y, g, X; kwargs...) = copyto!(Y, X)
```
This should be easier to handle for the compiler.

src/groups/semidirect_product_group.jl
```diff
@@ -272,6 +272,130 @@ function _compose!(
...
+    SDPG::LieGroup{𝔽,<:LeftSemidirectProductGroupOperation,<:ProductManifold}, g, h, X
+) where {𝔽}
+
+"$(_doc_LSDP_diff_left_compose)"
+function diff_left_compose!(
+    SDPG::LieGroup{𝔽,<:LeftSemidirectProductGroupOperation,<:ProductManifold}, Y, g, h, X
+) where {𝔽}
+    PM = SDPG.manifold
+    G, H = map(LieGroup, PM.manifolds, SDPG.op.operations)
+    A = GroupAction(SDPG.op.action_type, G, H)
```
mateuszbaran, line 305: I think we should split this out to a separate function to reduce code duplication. It's already in 4 other places and I suspect they don't all get the order of G and H correct (it changes between left and right semidirect product).

src/groups/semidirect_product_group.jl
mateuszbaran: You're welcome :). The math looks OK but I didn't check it thoroughly.

mateuszbaran:
Could you also add a test that the new `diff_left_compose` gives for SE(3) as a semidirect product group the same result as `diff_left_compose` for SE(3) treated as a subgroup of GL(4) (which should use the `diff_left_compose` for matrix multiplication groups) ?

src/groups/special_orthogonal_group.jl
kellertuer: I always chose the wrong one of these two. consistently. ;)

src/groups/semidirect_product_group.jl
kellertuer: do you have a good idea for a name?

kellertuer: concerning the order - for me this takes a lot of thinking and scribbling a few formulae somewhere to implement these, so I hope they are in the right order.

mateuszbaran: Maybe `semidirect_group_action(SPDG)` ?

The order of G and H is inconsistent between different methods so I'm a bit suspicious. Anyway, refactoring it to a new function will fix any issues that might be there.

mateuszbaran:
Everything else looks fine, with that one test added and that minor refactoring (and passing CI) this should be ready to be merged.

mateuszbaran:
Oh, sorry, I missed the part about `adjoint_action`. I think there is not much reason to do Manifolds.jl archaeology there but maybe let's talk tomorrow about it during a break or before the conference opening?

kellertuer:
For now I just exchanged one error message with a next that I do not understand yet, so I can not say this is done, since it das still not work (even less for the default representation in homogeneous coordinates.

I can check for more consistency with left/right, but I think I mainly used G \ltimes H for left and H \rtimes G for right, so that G is always the “plain” one and H the one affected by the action.

kellertuer:
> Could you also add a test that the new diff_left_compose gives for SE(3) as a semidirect product group the same result as diff_left_compose for SE(3) treated as a subgroup of GL(4) (which should use the diff_left_compose for matrix multiplication groups) ?

good idea for a test assuming that we have diff left compose on GL(n). Otherwise this PR slowly gets a bit too many things to fix.

mateuszbaran:
> good idea for a test assuming that we have diff left compose on GL(n). Otherwise this PR slowly gets a bit too many things to fix.

Yes, we do have it. I think I will try adding it myself.

mateuszbaran:
BTW, what's the recommended way of doing that conversion? In Manifolds.jl I'd do embed and project with `SpecialEuclideanInGeneralLinear`.

mateuszbaran:
Uh, the test failure with diff_left_compose! on generic semidirect product groups is too complicated for me I'm afraid. I could do it for SE(n) specifically but the generic code would take me hours to properly understand to be able to fix it and I have too much other stuff to do.

kellertuer:
Then let's doo that for SE(n) first and do the rest later. I mainly have not yet gotten around to see how apply_diff would look in general.

So feel free to add the SE(n) case and then we see what we do with the generic code I came up with today, If I can not test it properly I'll remove it again.

src/groups/semidirect_product_group.jl
kellertuer: Will do that, maybe during the day if I find a. bit of time. I am tending a bit towards a name more like `semidifrect_group_elements` or so since it returns more that just the action.

mateuszbaran:
OK, I can do SE(n). Could you just show me the intended way to convert between the `ArrayPartition` representation and GL(n+1) embedding for SE(n)?

kellertuer:
I think we are nearly there. Currently the test for the `Y` we compute fails, but if I `pullback` that (once more than we already do in the code) this is actually correct.

Have to check where that comes from.

kellertuer:
...or it might be that on rotations we represent also tangent vectors in the LieAlgebra and either `push_forward` or `pull_back` do not take that into account correctly.

kellertuer:
Found it. So now there is only missing, I added them to the initial post.

@Affie do you somehow want to give this a try already? The new retraction type allows you now to use any retraction on the base manifold. The complicated part was to get the “tangent adaption” right, i.e. to move from Lie algebra(s) to the tangent spaces (or the inverse).

src/groups/semidirect_product_group.jl
mateuszbaran: OK, `semidirect_group_elements` sounds good too.

kellertuer:
I noticed, someone (tm) already wrote all the conversions we need ;)

I wrote the first test that I think you meant:

https://github.com/JuliaManifolds/LieGroups.jl/blob/a56ce8e4eebdfef7659d9f6792b6b541ac00e0d9/test/groups/test_special_euclidean_group.jl#L379-L385

but it does something different, can you check it is what you meant to check.

mateuszbaran (approved):

mateuszbaran:
I've tried improving coverage but now we have weird failures :disappointed:

kellertuer:
Not too weird, for me it looks like we forgot in the convert to set the zero entries explicitly to zero, especially when coming along with memory, they just seem to stay uninitialised?

kellertuer:
I think this should be ready. Tests pass, code coverage is there, documentation is fine.

docs/Project.toml
```diff
@@ -10,3 +10,6 @@ Documenter = "1"
 DocumenterCitations = "1.3"
 DocumenterInterLinks = "1"
 LinearAlgebra = "1.11"
+
+[sources]
+LieGroups = { path = ".." }
```
mateuszbaran, line 15: What does this thing do?

mateuszbaran (approved):
LGTM, up to not knowing what the `sources` section is useful for here.

docs/Project.toml
kellertuer: Formerly we manually said that we add LieGroups.jl in development mode from the parent folder.

This is a bit easier and available since 1.11, and states that LieGroups.jl should be loaded from there. That way we do not have to do that anew manually every run.

kellertuer: cf. https://github.com/JuliaLang/Pkg.jl/pull/3783#issuecomment-1935523553 or you can also easier add unregistered packages https://julialang.org/blog/2024/10/julia-1.11-highlights/#sources_section_in_projecttoml_in_pkgjl.

But here it is just an easier path to always load LieGroups.jl from parent folder

mateuszbaran: OK, so it seems to be mostly relevant to building docs, right?

kellertuer: yes, it simplifies the code in make.jl a bit – I mean its also the project.toml for the docs ;)

PR #48, opened 2025-06-22
Fix #44

### LINKED ISSUE #44: Numerical issue with exp/log on SE with small angle
For a small angle in exp/log the results become numerically unstable. 

Maybe it is because `α ≈ 0` is equivalent to `α == 0` here: https://github.com/JuliaManifolds/LieGroups.jl/blob/859815f1d7352ef2c5e7e79b637f453abc6c92af/src/groups/special_euclidean_group.jl#L688

It can maybe be fixed with something like `isapprox(α, 0; atol=1e-8)`

Here is an example:

```julia
julia> G = SpecialEuclideanGroup(2)
SpecialEuclideanGroup(2)

julia> X = hat(LieAlgebra(G), [1e-8, 1, 0])
3×3 Matrix{Float64}:
 0.0     -1.0e-8  1.0
 1.0e-8   0.0     0.0
 0.0      0.0     0.0

julia> p = exp(G, X)
3×3 Matrix{Float64}:
 1.0     -1.0e-8  1.0
 1.0e-8   1.0     0.0
 0.0      0.0     1.0

julia> X = log(G, p)
3×3 Matrix{Float64}:
 0.0     -1.0e-8   Inf
 1.0e-8   0.0     NaN
 0.0      0.0       0.0
```

mateuszbaran:
I've basically copied the old code from Manifolds.jl which works well.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,7 +5,7 @@ All notable Changes to the Julia package `LieGroups.jl` will be documented in th
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [0.1.2] unreleased
+## [0.1.2] 2025-06-22
 
 ### Added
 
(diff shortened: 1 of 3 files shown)


kellertuer (approved):
lgtm

mateuszbaran:
Should we add it to the 0.1.2 release?

kellertuer:
Sure.

Issue #45, opened 2025-06-19
Support StaticArrays


Affie:
In Manifolds.jl we used SaticArrays for computations and as far as I remember it mostly propagated correctly for the manifolds we used.

For example, a SMatrix here is converted to a MMatrix:
```julia
julia> p = exp(G, SMatrix{3, 3, Float64}(I))
3×3 MMatrix{3, 3, Float64, 9} with indices SOneTo(3)×SOneTo(3):
 1.0  -0.0  0.0
 0.0   1.0  0.0
 0.0   0.0  1.0
```
or 

```julia
julia> T = ArrayPartition{Float64, Tuple{SVector{2, Int64}, SVector{2, Float64}}}
ArrayPartition{Float64, Tuple{SVector{2, Int64}, SVector{2, Float64}}}

julia> X = hat(LieAlgebra(G), SA[1, 0, 0.01], T)
([1.0, 0.0], [0.0 -0.01; 0.01 0.0])

julia> typeof(X)
ArrayPartition{Float64, Tuple{Vector{Float64}, Matrix{Float64}}}
```

Is it still possible to support StaticArrays? 
Is it worth it, we previously got a huge speed improvement after Mateusz recommended it (we use vectors of StaticArray). Should we revisit our decision?

We can help port the code from Manifolds.jl or implement on a case by case basis. 

Then concerning the T::Type parameter in some of the functions such as hat.  We don't write specialized code for every group, so I don't currently know how to provide the T parameter to the functions. 
I'm thinking of adding a trait such as `default_type(G::SE2) =  ArrayPartition{Float64, Tuple{SVector{2, Int64}, SVector{2, Float64}}}`
We don't have a common place to put it though, so would have to create a new package. 
What do you think, or what would you suggest? Does LieGroups have a place for something like this?

kellertuer:
That should be possible the same way, I would be surprised if it fails in many places, but there might be some.

Probably @mateuszbaran is the best to check with, my experience with StaticArrays is very limited;

Such code would probably be put into an extension.

mateuszbaran:
Sure, I can add `StaticArrays` methods in an extension here. Everything should work fine like it did in Manifolds.jl eventually. Adding those dispatches just wasn't a high priority. I will try to find some time next week to copy group-related StaticArrays.jl stuff from Manifolds.jl to LieGroups.jl.

mateuszbaran:
[...]

PR #49, opened 2025-06-23
StaticArrays.jl specializations

mateuszbaran:
Aimed at resolving #45 .

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * New StaticArrays.jl specializations for multiple functions, including:
   * `exp` on the orthogonal and special orthogonal group in 2 and 3 dimensions.
+  * `get_coordinates` on the orthogonal and special orthogonal group in 2 and 3 dimensions.
 
 ## [0.1.2] 2025-06-22
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,22 +5,17 @@ All notable Changes to the Julia package `LieGroups.jl` will be documented in th
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [0.1.3] unreleased
-
-### Added
-
-* New StaticArrays.jl specializations for multiple functions, including:
-  * `exp` and `log` on the orthogonal and special orthogonal group in 2 and 3 dimensions.
-  * `get_coordinates` and `get_vector` on the orthogonal and special orthogonal group in 2 and 3 dimensions, for `LieAlgebraOrthogonalBasis`.
-* More generic implementation of non-mutating `get_vector_lie` on `AbstractProductGroupOperation` groups.
-
-## [0.1.2] 2025-06-23
+## [0.1.2] 2025-06-24
 
 ### Added
 
 * `is_flat` for `SpecialEuclideanGroup`
 * `inner` and `norm` for `LieAlgebra` to compute the inner product and norm on the Lie algebra.
 * a test suite function for `identity_element`.
+* New StaticArrays.jl specializations for multiple functions, including:
+  * `exp` and `log` on the orthogonal and special orthogonal group in 2 and 3 dimensions.
+  * `get_coordinates` and `get_vector` on the orthogonal and special orthogonal group in 2 and 3 dimensions, for `LieAlgebraOrthogonalBasis`.
+* More generic implementation of non-mutating `get_vector_lie` on `AbstractProductGroupOperation` groups.
 
 ### Changed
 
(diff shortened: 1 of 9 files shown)


test/groups/test_circle_group.jl
```diff
@@ -3,6 +3,8 @@ using LieGroups, Random, Test
 using ManifoldsBase: ℂ, ℝ
 using Manifolds: Circle, Sphere
 
+using StaticArrays
```
kellertuer, line 6: For now StaticArrays is a dependency for Manifolds.jl anyways right? then this can be that as well, for sure; in general one could think (later) about maybe moving it to an ext?

mateuszbaran: Yes, it's a direct dependency of Manifolds.jl. I've considered moving it to an extension but it's a lot of work to do it without introducing performance regressions. And IIRC it would still be an indirect dependency so no load time reduction anyway.

kellertuer: here I do agree, we would have to start with that over in manifolds if we ever want to do that. I am also not 100% sure its worth it, I was just curious.

mateuszbaran: It might be worth doing at some point in the future but not right now.

kellertuer: Sure, definitely not urgent.

mateuszbaran:
I think this covers everything we have in Manifolds.jl. @Affie this seems to cover your examples but please keep reporting things that don't work as expected.

NEWS.md
```diff
@@ -5,6 +5,15 @@ All notable Changes to the Julia package `LieGroups.jl` will be documented in th
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.1.3] unreleased
```
kellertuer: We never released 0.1.2, so this can just be part of that as well (I am not sure who added the date to 0.1.2)

test/groups/test_orthogonal_group.jl
```diff
@@ -63,6 +94,28 @@ using LinearAlgebra: I
     @testset "O(3) special cases" begin
         @test is_identity(H, exp(H, zeros(3, 3)))
     end
+
+    @testset "StaticArrays specializations for O(3)" begin
+        ps = SMatrix{3,3,Float64}(I)
+        Xs = SA[0.0 0.1 0.0; -0.1 0.0 0.0; 0.0 0.0 0.0]
+        qs = exp(H, ps, Xs)
+
+        q_ref = SA[
```
kellertuer, line 103: I would prefer to avoid tests with “magic numbers” if possible, since I am not sure where these are from, what to do when the test (109) fails? (besides asking you ;))

kellertuer (approved):
I'll already mark this as OK, but please check the version number – and check whether you can get rid of magic numbers.

test/groups/test_orthogonal_group.jl
mateuszbaran: Sorry, those tests should be for approximate equality -- I'll fix that. This is the expected outcome. What would you suggest to replace this number with?

NEWS.md
mateuszbaran: OK, I've merged the the release notes.

test/groups/test_orthogonal_group.jl
kellertuer: To me it is a bit unclear _why_ that is the outcome. It is a bit like “trust me these magic numbers are right.”
And they are not as obvious as a manifold dimension or so where I feel testing against a number is fine.

So if this test ever fails – all I can do is reach out to you, I have zero chance to fix this myself. Nor has anyone else. The numbers just magically appear here.
(That is not as worse as random tests, sure).

So if you have an idea to avoid such a point that would be great – even if it is only in the long run and not directly now on this PR.

mateuszbaran: You can always check the expected outcome yourself. It's not particularly difficult and I'm not needed for that. I don't know any better way to handle this.

kellertuer: I usually prefer and alternative path to compute it, instead of our exp, the matrix exp or something like that, so that is is clear where it comes from.
I mean you have the magic numbers also from some function call. I am quite sure you did not compute that precision by hand with pen & paper ;)

mateuszbaran: Yes, I took it from the output; I wouldn't make a separate implementation from scratch either. That's what I only do for complicated stuff and in limited scope. Anyway, a separate implementation suffers from the exact same problem: if the test fails, you have to figure out where the error is. You just have two implementations available at the same time instead of at two different commits. And if some bug in a dependency affects both computations in the same way, it could go unnoticed.

kellertuer: It would still be easier to follow than magic numbers. But let's maybe address that in a future PR.

mateuszbaran: I can't recall a single issue caused by specifically a magic number in a test like this. So it feels to me like wasting time fixing issues that don't exist. But thanks for agreeing that I don't have to waste time right now.

Issue #54, opened 2025-07-16
ValidationLieGroup Method errors: possible typo and submanifold_components


Affie:
It looks like there is a typo in "validation_group.jl" with `widthin` as I get this error:

```julia
ERROR: MethodError: no method matching isapprox(::Float64, ::Int64; widthin::typeof(log))
This error has been manually thrown, explicitly, so the method may exist but be intentionally marked as unimplemented.

Closest candidates are:
  isapprox(::Number, ::Number; atol, rtol, nans, norm) got unsupported keyword argument "widthin"
   @ Base floatfuncs.jl:220
  isapprox(::Missing, ::Any; kwargs...)
   @ Base missing.jl:90
  isapprox(::Any, ::Missing; kwargs...)
   @ Base missing.jl:91
  ...

Stacktrace:
  [1] kwerr(::@NamedTuple{widthin::typeof(log)}, ::Function, ::Float64, ::Int64)
    @ Base ./error.jl:165
  [2] check_point(M::Rotations{…}, p::MMatrix{…}; kwargs::@Kwargs{…})
    @ Manifolds ~/.julia/packages/Manifolds/ksguI/src/manifolds/GeneralUnitaryMatrices.jl:91
  [3] _check_point(G::LieGroup{…}, Rotn::Rotations{…}, Rn::Euclidean{…}, op1::MatrixMultiplicationGroupOperation, op2::AdditionGroupOperation, g::ArrayPartition{…}; kwargs::@Kwargs{…})
    @ LieGroups ~/.julia/packages/LieGroups/4KYR9/src/groups/special_euclidean_group.jl:219
  [4] check_point(G::LieGroup{…}, g::ArrayPartition{…}; kwargs::@Kwargs{…})
    @ LieGroups ~/.julia/packages/LieGroups/4KYR9/src/groups/special_euclidean_group.jl:208
  [5] is_point(M::LieGroup{…}, p::ArrayPartition{…}; error::Symbol, kwargs::@Kwargs{…})
    @ ManifoldsBase ~/.julia/packages/ManifoldsBase/nqeOA/src/ManifoldsBase.jl:809
  [6] is_point
    @ ~/.julia/packages/ManifoldsBase/nqeOA/src/ManifoldsBase.jl:795 [inlined]
  [7] #is_point#167
    @ ~/.julia/packages/LieGroups/4KYR9/src/groups/validation_group.jl:491 [inlined]
  [8] is_point
[...]

PR #55, opened 2025-07-16
Fix a typo

kellertuer:
`widthin` should be `within`. Adresses Part 2/3 from #54.


Affie (approved):

kellertuer:
I will wait a bit after merging this with registering a new version (already added the news, but did not yet bump version), because I think the remaining error is on the how do we pass `get_vector` around here internally.” side.

PR #58, opened 2025-07-30
Add the lie_bracket for the SpecialEuclideanGroup

### LINKED ISSUE #57: `lie_bracket` possibly not implemented for semidirect products: throws method error for SE(n)
```julia
G = SpecialEuclideanGroup(2; variant = :right)
# also the same error on these:
# G = SpecialEuclideanGroup(2)
# G = SpecialEuclideanGroup(3; variant = :right)
# G = SpecialEuclideanGroup(3)
X = zero_vector(LieAlgebra(G), ArrayPartition)
# also the same error on non ArrayPartition
# X = zero_vector(LieAlgebra(G))

julia> LieGroups.lie_bracket(LieAlgebra(G), X, X) 
ERROR: MethodError: no method matching lie_bracket!(::Fiber{…}, ::ArrayPartition{…}, ::ArrayPartition{…}, ::ArrayPartition{…})
The function `lie_bracket!` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  lie_bracket!(::Fiber{𝔽, ManifoldsBase.TangentSpaceType, <:LieGroup{𝔽, Op, M}, LieGroups.Identity{Op}}, ::Any, ::Any, ::Any) where {𝔽, Op<:ProductGroupOperation, M<:ProductManifold}
   @ LieGroups ~/.julia/dev/LieGroups/src/groups/product_group.jl:300
  lie_bracket!(::Fiber{𝔽, ManifoldsBase.TangentSpaceType, <:LieGroup{𝔽, Op, M}, LieGroups.Identity{Op}}, ::Any, ::Any, ::Any) where {𝔽, Op<:PowerGroupOperation, M<:AbstractPowerManifold}
   @ LieGroups ~/.julia/dev/LieGroups/src/groups/power_group.jl:299
  lie_bracket!(::Fiber{𝔽, ManifoldsBase.TangentSpaceType, <:LieGroup{𝔾, O, M} where M<:AbstractManifold{𝔾}, LieGroups.Identity{O}}, ::Any, ::Any, ::Any) where {𝔽, 𝔾, O<:AbelianMultiplicationGroupOperation}
   @ LieGroups ~/.julia/dev/LieGroups/src/group_operations/multiplication_operation_abelian.jl:336
  ...

Stacktrace:
 [1] lie_bracket(𝔤::Fiber{…}, X::ArrayPartition{…}, Y::ArrayPartition{…})
   @ LieGroups ~/.julia/dev/LieGroups/src/Lie_algebra/Lie_algebra_interface.jl:295
 [2] top-level scope
   @ REPL[33]:1
Some type information was truncated. Use `show(err)` to see complete types.
```

Affie:
To close #57 
Moved [lie_bracket from Manifolds.jl](https://github.com/JuliaManifolds/Manifolds.jl/blob/db305e085ae5608310c10a6d19e3524a118b6e81/src/groups/special_euclidean.jl#L804-L825) and updated to fit with LieGroups.jl 

- [x] Needs to be tested
- [x] Fix documentation
- [x] ~Improve duplications: the matrix version is currently duplication of the `lie_bracket` on `MatrixMultiplicationGroupOperation`~

### DIFF
--- a/ext/LieGroupsRecursiveArrayToolsExt/special_euclidean_group_RAT_ext.jl
+++ b/ext/LieGroupsRecursiveArrayToolsExt/special_euclidean_group_RAT_ext.jl
@@ -186,6 +186,37 @@ function LinearAlgebra.norm(
     return LinearAlgebra.norm([n1, n2])
 end
 
+"""
+    lie_bracket(𝔰𝔢::LieAlgebra{ℝ, SpecialEuclideanGroupOperation, SpecialEuclidean}, X::ArrayPartition, Y::ArrayPartition)
+    lie_bracket!(𝔰𝔢::LieAlgebra{ℝ, SpecialEuclideanGroupOperation, SpecialEuclidean}, Z::ArrayPartition, X::ArrayPartition, Y::ArrayPartition)
+
+Calculate the Lie bracket between elements `X` and `Y` of the special Euclidean Lie
+algebra. For the matrix representation the formula is ``[X, Y] = XY-YX``, while in the `ArrayPartition` representation the
+formula reads ``[X, Y] = [(X_t, X_R), (Y_t, Y_R)] = (X_R * Y_t - Y_R * X_t, X_R * Y_R - Y_R * X_R)``.
+"""
+function LieGroups.lie_bracket!(
+    𝔤::LieGroups.LieAlgebra{
+        ℝ,
+        <:LieGroups.SpecialEuclideanGroupOperation,
+        <:LieGroups.SpecialEuclideanGroup,
+    },
+    Z::ArrayPartition,
+    X::ArrayPartition,
+    Y::ArrayPartition,
+)
+    G = LieGroups.base_lie_group(𝔤)
+    SOn, _ = LieGroups._SOn_and_Tn(G)
+    X_t = submanifold_components(G, X, :Translation)
+    X_R = submanifold_components(G, X, :Rotation)
+    Y_t = submanifold_components(G, Y, :Translation)
+    Y_R = submanifold_components(G, Y, :Rotation)
+    Z_t = submanifold_components(G, Z, :Translation)
+    Z_R = submanifold_components(G, Z, :Rotation)
+    LieGroups.lie_bracket!(LieAlgebra(SOn), Z_R, X_R, Y_R)
+    Z_t .= X_R * Y_t .- Y_R * X_t
+    return Z
+end
+
 function ManifoldsBase.submanifold_component(
     G::LieGroups.LeftSpecialEuclideanGroup,
     g::Union{ArrayPartition,SpecialEuclideanProductPoint},
--- a/ext/LieGroupsRecursiveArrayToolsExt/special_euclidean_group_RAT_ext.jl
+++ b/ext/LieGroupsRecursiveArrayToolsExt/special_euclidean_group_RAT_ext.jl
@@ -196,9 +196,7 @@ formula reads ``[X, Y] = [(X_t, X_R), (Y_t, Y_R)] = (X_R * Y_t - Y_R * X_t, X_R 
 """
 function LieGroups.lie_bracket!(
     𝔤::LieGroups.LieAlgebra{
-        ℝ,
-        <:LieGroups.SpecialEuclideanGroupOperation,
-        <:LieGroups.SpecialEuclideanGroup,
+        ℝ,<:LieGroups.SpecialEuclideanGroupOperation,<:LieGroups.SpecialEuclideanGroup
     },
     Z::ArrayPartition,
     X::ArrayPartition,
--- a/ext/LieGroupsRecursiveArrayToolsExt/special_euclidean_group_RAT_ext.jl
+++ b/ext/LieGroupsRecursiveArrayToolsExt/special_euclidean_group_RAT_ext.jl
@@ -204,12 +204,12 @@ function LieGroups.lie_bracket!(
 )
     G = LieGroups.base_lie_group(𝔤)
     SOn, _ = LieGroups._SOn_and_Tn(G)
-    X_t = submanifold_components(G, X, :Translation)
-    X_R = submanifold_components(G, X, :Rotation)
-    Y_t = submanifold_components(G, Y, :Translation)
-    Y_R = submanifold_components(G, Y, :Rotation)
-    Z_t = submanifold_components(G, Z, :Translation)
-    Z_R = submanifold_components(G, Z, :Rotation)
+    X_t = submanifold_component(G, X, :Translation)
+    X_R = submanifold_component(G, X, :Rotation)
+    Y_t = submanifold_component(G, Y, :Translation)
+    Y_R = submanifold_component(G, Y, :Rotation)
+    Z_t = submanifold_component(G, Z, :Translation)
+    Z_R = submanifold_component(G, Z, :Rotation)
     LieGroups.lie_bracket!(LieAlgebra(SOn), Z_R, X_R, Y_R)
     Z_t .= X_R * Y_t .- Y_R * X_t
     return Z
--- a/ext/LieGroupsRecursiveArrayToolsExt/special_euclidean_group_RAT_ext.jl
+++ b/ext/LieGroupsRecursiveArrayToolsExt/special_euclidean_group_RAT_ext.jl
@@ -186,14 +186,34 @@ function LinearAlgebra.norm(
     return LinearAlgebra.norm([n1, n2])
 end
 
-"""
-    lie_bracket(𝔰𝔢::LieAlgebra{ℝ, SpecialEuclideanGroupOperation, SpecialEuclidean}, X::ArrayPartition, Y::ArrayPartition)
-    lie_bracket!(𝔰𝔢::LieAlgebra{ℝ, SpecialEuclideanGroupOperation, SpecialEuclidean}, Z::ArrayPartition, X::ArrayPartition, Y::ArrayPartition)
+_doc_lie_bracket_SEn_RAT = """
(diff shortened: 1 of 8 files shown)


test/groups/test_special_euclidean_group.jl
```diff
@@ -18,6 +18,7 @@ using StaticArrays
         inv,
         is_flat,
         is_identity,
+        lie_bracket,
```
Affie, line 21: I tried to add this to enable testing, ~but unrelated tests seem to break now~ for some reason some tests broke locally but that is no longer the case. Stil need to figure out how the testing works.

ext/LieGroupsRecursiveArrayToolsExt/special_euclidean_group_RAT_ext.jl
```diff
@@ -186,6 +186,35 @@ function LinearAlgebra.norm(
...
+    X_R = submanifold_components(G, X, :Rotation)
+    Y_t = submanifold_components(G, Y, :Translation)
+    Y_R = submanifold_components(G, Y, :Rotation)
+    Z_t = submanifold_components(G, Z, :Translation)
+    Z_R = submanifold_components(G, Z, :Rotation)
+    LieGroups.lie_bracket!(LieAlgebra(SOn), Z_R, X_R, Y_R)
+    Z_t .= X_R * Y_t .- Y_R * X_t
+    return Z
+end
+
```
Affie: This is from https://github.com/JuliaManifolds/Manifolds.jl/blob/db305e085ae5608310c10a6d19e3524a118b6e81/src/groups/special_euclidean.jl#L804-L825 but I tried to update to match both the left and right variants

src/groups/special_euclidean_group.jl
```diff
@@ -777,6 +777,17 @@ function LinearAlgebra.norm(
...
+function lie_bracket!(
+    ::LieAlgebra{ℝ,<:SpecialEuclideanGroupOperation,<:SpecialEuclideanGroup},
+    Z::AbstractMatrix,
+    X::AbstractMatrix,
+    Y::AbstractMatrix,
+)
+    mul!(Z, X, Y)
+    mul!(Z, Y, X, -1, true)
+    return Z
+end
```
Affie: If I include `SpecialEuclideanMatrixTangentVector` then `mul!` fails.
So, I don't know what the preferred way to deal with `SpecialEuclideanMatrixTangentVector` is here.

test/groups/test_special_euclidean_group.jl
kellertuer: This should be enough to start testing the function, since it activates to test https://github.com/JuliaManifolds/LieGroups.jl/blob/61f6053fcc7e3dd8091f3f1efde12c3a72bf7165/test/LieGroupsTestSuite.jl/LieGroupsTestSuite.jl#L855-L877 within the test suite.

src/groups/special_euclidean_group.jl
kellertuer: One could specialise not for the `SpecialEuclideanMatrixTangentVector` that basically cals this existing function with the `X.value, Y.value` and works in-place of `Z.value`.

Affie: I extended `Base.:*` for `SpecialEuclideanMatrixTangentVector` to see if it now works. 
And changed mul! to *

test/groups/test_special_euclidean_group.jl
kellertuer: Code coverage already looks good!

The test suite is really meant to make it that easy: You specify which functions to test and then certain properties are checked (minimum: That allocating and in-place variant do the same).

Affie: I was just looking at some of the test suite stuff now and I think I get the idea. It looks good.

Affie:
I think this is as far as my current knowledge of LieGroups.jl can take me.

test/groups/test_special_euclidean_group.jl
Affie: I was wondering, if we can know a group is a matrix lie group, we can expand the tests to include the definition and properties, something like:
```julia
   # .... @ line 876 above
        if X isa AbstractMatrix && Y isa AbstractMatrix
            # Assume a Matrix Lie group for a AbstractMatrix Tangent Vector:
            #test against the matrix commutator definition
            @test isapprox(𝔤, Z1, X * Y - Y * X)
            #test anti-symmetry property
            Z2 = lie_bracket(𝔤, Y, X)
            @test isapprox(𝔤, Z1, -Z2)
        end
```

src/groups/special_euclidean_group.jl
kellertuer: For that we should check with @mateuszbaran because I think that would allocate too much and I would prefer to still use `mul!` for them as well, but you could also extend `Base.mul!` here then.

test/groups/test_special_euclidean_group.jl
kellertuer: Nice that you like the test suite, I put a reasonable amount of thinking into that.

For the proposed check the function has an `expected=` field to provide an expected value.
Checking for the types with a `isa` seems a bit too “hardcoded’ for me. But you can pass that expected value into the test suite as well.

kellertuer:
This already looks very good – we have code coverage already and Documentation is rendering!

We should check for efficiency with the special type and I might change the doc string slightly, making the long formula maybe an offset one and explaining the `_R` and `_t` part a bit, but I can do that later today, I hope.

src/groups/special_euclidean_group.jl
Affie: I was a bit worried about `mul!`, from the docs: _"mul!(Y, A, B) -> Y ... Note that Y must not be aliased with either A or B."_
It might therefore be a problem if someone calls `lie_bracket!(G, X, X, Y)`, but I don't know if that is something that will be done or if it can maybe just be prevented with a similar comment in the docs.

kellertuer: You are right, but this uncarefulness we also have for the “normal” matrix variant for now – for historical reasons, since the old one had that as well. But we should maybe be more careful there then.

src/groups/special_euclidean_group.jl
```diff
@@ -777,6 +777,25 @@ function LinearAlgebra.norm(
...
+function lie_bracket!(
+    ::LieAlgebra{ℝ,<:SpecialEuclideanGroupOperation,<:SpecialEuclideanGroup},
+    Z::Union{AbstractMatrix,SpecialEuclideanMatrixTangentVector},
+    X::Union{AbstractMatrix,SpecialEuclideanMatrixTangentVector},
+    Y::Union{AbstractMatrix,SpecialEuclideanMatrixTangentVector},
+)
+    # FIXME mul! does not work with SpecialEuclideanMatrixTangentVector
+    # mul!(Z, X, Y)
+    # mul!(Z, Y, X, -1, true)
+    Z .= X * Y - Y * X
```
kellertuer, line 812: Maybe you are indeed right that we have to write this exactly like this to avoid side effects.

kellertuer:
I just reworked the docs and hopefully made them nicer.

src/groups/special_euclidean_group.jl
Affie: I see in a different part of the code base you used: `(k === g || k === h) ? copyto!(k, g * h) : mul!(k, g, h)`

kellertuer: ah we could also do something like that, sure.

Affie:
@kellertuer, what do we still need to finish this PR? The only things that I can think of is:

-  the duplication where both these use the same formula of `XY-YX`
`lie_bracket!(::LieAlgebra{ℝ,<:SpecialEuclideanGroupOperation,<:SpecialEuclideanGroup}...` and 
`lie_bracket!(::LieAlgebra{𝔽,O,<:LieGroup{𝔽,O}}, ...) where {𝔽,O<:MatrixMultiplicationGroupOperation}`
Maybe that is ok for now?

- Safety of the `MatrixMultiplicationGroupOperation` version where we can maybe add as discussed:
```julia
    # Avoid issues if Z is aliased with either X or Y
    if Z === X || Z === Y
        Z .= X * Y - Y * X
    else
        mul!(Z, X, Y)
        mul!(Z, Y, X, -1, true)
    end
```
It's not decisions I feel I can make, if you tell me what to to I can help, otherwise I leave it up to you.

kellertuer:
Why should we not finish this PR? It seems to add a function that was present in Manifolds.jl, so it should be here as well – I just struggle a bit keeping track of all the (by now 10+) PRs I am contributing to in parallel ;)

For the duplication, that is actually fine with me.

For the `mul!` vs `.=`/`*` – the simpler code would just be the `.=` line, but is we have tests coverage for both cases, I am also fine with your proposal.

Affie:
> Why should we not finish this PR? It seems to add a function that was present in Manifolds.jl, so it should be here as well

Sorry, you misunderstood, I was asking what was still outstanding.

> For the duplication, that is actually fine with me.

Fine with me as well

> For the mul! vs .=/* – the simpler code would just be the .= line, but is we have tests coverage for both cases, I am also fine with your proposal.

I also prefer the simpler code.

I think this is ready then?

kellertuer (approved):
This looks good to me, let's maybe wait a day or two whether someone else has something to add – then I can merge it.

Thanks for contributing!

mateuszbaran:
Thanks for the contribution!

> * Safety of the `MatrixMultiplicationGroupOperation` version where we can maybe add as discussed:

I'm fine with just having the simpler code but there is a safe way to handle this, using `Base.mightalias` instead of `===`. We generally don't use that function much in JuliaManifolds because we don't work much with cases where it's actually a performance bottleneck but we surely could do that if needed.

mateuszbaran (approved):

PR #59, opened 2025-07-30
Fix RightGroupOperationAction bug

Affie:
From doc (and my current understanding) it looks like `RightGroupOperationAction` should be a `AbstractRightGroupActionType`.


kellertuer:
Well-spotted – yo are correct, that seems to come from some copy paste.
Thanks for the fix. Could you also add a small item to the News.md for completeness?

(I specifically added this one CI checking for that so I get reminded to have that on every PR).

The rest looks fine to me.

kellertuer (approved):

Issue #63, opened 2025-08-13
Issue with `exp!` when `h === g` on SpecialOrthogonalGroup


Affie:
`exp!(G, h, g, X)` does not work correctly if `h === g`
```julia
julia> M = SpecialOrthogonalGroup(3)
SpecialOrthogonalGroup(3)

julia> p = identity_element(M)
3×3 Matrix{Float64}:
 1.0  0.0  0.0
 0.0  1.0  0.0
 0.0  0.0  1.0

julia> X = hat(LieAlgebra(M), [0.1,0,0])
3×3 Matrix{Float64}:
  0.0  -0.0   0.0
  0.0   0.0  -0.1
 -0.0   0.1   0.0

julia> exp!(M, p, p, X)
3×3 Matrix{Float64}:
 1.0  0.0        0.0
 0.0  0.980067  -0.198669
 0.0  0.198669   0.980067

julia> p
3×3 Matrix{Float64}:
 1.0  0.0        0.0
 0.0  0.980067  -0.198669
 0.0  0.198669   0.980067

julia> exp!(M, p, p, X)
3×3 Matrix{Float64}:
 1.0  0.0        0.0
 0.0  0.980067  -0.198669
 0.0  0.198669   0.980067

julia> p
3×3 Matrix{Float64}:
 1.0  0.0        0.0
 0.0  0.980067  -0.198669
 0.0  0.198669   0.980067
```

Affie:
It looks like  a lot of other groups have similar issues. If I add [this test](https://github.com/JuliaManifolds/LieGroups.jl/pull/64) I get 47 failures.

kellertuer:
I might be a bit blind here? To me all 4 last matrices are the same. Is that what you expect? Should that be different? I am not yet following completely (maybe since I am before first coffee).

In general I think we were not always that strict in “aliasing is allowed”, cf. the recent thing were we discussed whether or not to use [`mul!`](https://docs.julialang.org/en/v1/stdlib/LinearAlgebra/#LinearAlgebra.mul!) where that is not allowed. I think we can work towards that being allowed – but it might take some time.

Affie:
> In general I think we were not always that strict in “aliasing is allowed”

I remember an issue I had with Manopt not converging to the correct value due to `aliasing should be allowed`, I can't find it new though. I found this bug trying to track down new numerical failures in tests after switching to LieGroups.jl.

Affie:
[...]

PR #64, opened 2025-08-13
Test and fix exp! issue from #63

Affie:
Add a test for the `exp!` issue when `g === h`.

### DIFF
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -429,8 +429,8 @@ ManifoldsBase.exp(G::AbstractLieGroup, ::Any, ::Any)
 @doc "$_doc_exp"
 function ManifoldsBase.exp!(G::AbstractLieGroup, h, g, X)
     if Base.mightalias(g, h)
-       compose!(G, h, g, exp(G, X))
-       return h
+        compose!(G, h, g, exp(G, X))
+        return h
     else
         exp!(G, h, X)
         compose!(G, h, g, h)
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -428,9 +428,14 @@ ManifoldsBase.exp(G::AbstractLieGroup, ::Any, ::Any)
 
 @doc "$_doc_exp"
 function ManifoldsBase.exp!(G::AbstractLieGroup, h, g, X)
-    exp!(G, h, X)
-    compose!(G, h, g, h)
-    return h
+    if Base.mightalias(g, h)
+       compose!(G, h, g, exp(G, X))
+       return h
+    else
+        exp!(G, h, X)
+        compose!(G, h, g, h)
+        return h
+    end
 end
 
 _doc_exponential = """
--- a/test/LieGroupsTestSuite.jl/LieGroupsTestSuite.jl
+++ b/test/LieGroupsTestSuite.jl/LieGroupsTestSuite.jl
@@ -457,10 +457,16 @@ function test_exp_log(
             @test is_point(G, k1; error = :error, atol = atol)
             # exp
             k1 = exp(G, g, X)
+            l1 = exp(G, k1, X)
             if test_mutating
                 k2 = copy(G, g)
                 exp!(G, k2, g, X)
                 @test isapprox(G, k1, k2, atol = atol)
+                #test exp!(G, h, g, X) when h == g (issue #63)
+                l2 = copy(G, g)
+                exp!(G, l2, l2, X)
+                exp!(G, l2, l2, X)
+                @test isapprox(G, l1, l2, atol = atol)
             end
             @test is_point(G, k1; error = :error)
         end
--- a/.JuliaFormatter.toml
+++ b/.JuliaFormatter.toml
@@ -1,1 +1,0 @@
-style = "blue"
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,12 @@ All notable Changes to the Julia package `LieGroups.jl` will be documented in th
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.1.4] unreleased
+
+### Fixed
+
+* Fixed an issue, where `exp!(G, h, g, X)` would return a wrong result if the input `g`and the output `h`are aliased (#63).
+
 ## [0.1.3] 2025-08-04
 
 ### Added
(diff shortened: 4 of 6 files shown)


test/LieGroupsTestSuite.jl/LieGroupsTestSuite.jl
```diff
@@ -457,10 +457,16 @@ function test_exp_log(
...
             k1 = exp(G, g, X)
+            l1 = exp(G, k1, X)
             if test_mutating
                 k2 = copy(G, g)
                 exp!(G, k2, g, X)
                 @test isapprox(G, k1, k2, atol = atol)
+                #test exp!(G, h, g, X) when h == g (issue #63)
+                l2 = copy(G, g)
+                exp!(G, l2, l2, X)
+                exp!(G, l2, l2, X)
```
kellertuer, line 468: Why is line 467/468 double?

kellertuer (commented):
This is a good idea to test that, probably also for other functions where that might be an issue.

I would prefer to make that test optional similar to `test_mutating`. We can think about making it mandatory once all tests pass, but in the current form we would have to fix all current errors on this PR which might be a very long time to wait for until we find time to narrow down each and every error.

maybe something like `test_aliased` (idea taken from https://docs.julialang.org/en/v1/stdlib/LinearAlgebra/#LinearAlgebra.mul! in naming)?

Affie:
> I would prefer to make that test optional similar to test_mutating. We can think about making it mandatory once all tests pass, but in the current form we would have to fix all current errors on this PR which might be a very long time to wait for until we find time to narrow down each and every error. 

It looks like all except one of the failures are caused by this function:
https://github.com/JuliaManifolds/LieGroups.jl/blob/fe5a6f09478ae0c4b909db92b0f7004cc0e25057/src/interface.jl#L430-L434

test/LieGroupsTestSuite.jl/LieGroupsTestSuite.jl
Affie: It doesn't strictly have to be, but if g is an identity element calling the function the first time is correct, and it only fails after the second call.

kellertuer:
That would indicate it is compose for all of them? We could add an aliasing-test for `compose!` as well.

_edit:_ and it would explain why that previously not a problem, since that default we only could do here in `LieGroups.jl` I think and not that easily before.

Affie:
I don't know how to fix it without allocations, but just as a test I changed the `exp` to be allocating, so the issue is with `exp! ` changing the value of h and therefore g if g === h.
```julia
function ManifoldsBase.exp!(G::AbstractLieGroup, h, g, X)
       h = compose!(G, h, g, exp(G, X))
       return h
end
```

test/LieGroupsTestSuite.jl/LieGroupsTestSuite.jl
kellertuer: I still do not fully follow what this is testing then, sorry.

kellertuer:
I think we should make sure `compose!(G, k, g, h)` is able to handle aliased input `g===h`, `k===g` even `g===h===k`

test/LieGroupsTestSuite.jl/LieGroupsTestSuite.jl
kellertuer: Ah I now unterstood it based on the answer in 63. you expect the final `l2` to be the same as doing `exp(G, g, 2*X)` just as an in-place and iterated way.

src/interface.jl
```diff
@@ -428,9 +428,14 @@ ManifoldsBase.exp(G::AbstractLieGroup, ::Any, ::Any)
...
-    compose!(G, h, g, h)
-    return h
+    if Base.mightalias(g, h)
+       compose!(G, h, g, exp(G, X))
+       return h
+    else
+        exp!(G, h, X)
+        compose!(G, h, g, h)
+        return h
+    end
```
Affie: @kellertuer, what do you think of this as a fix? It fixes all groups aliasing except `HeisenbergGroup` which is implemented separately.

src/interface.jl
```diff
@@ -428,9 +428,14 @@ ManifoldsBase.exp(G::AbstractLieGroup, ::Any, ::Any)
 
 @doc "$_doc_exp"
 function ManifoldsBase.exp!(G::AbstractLieGroup, h, g, X)
-    exp!(G, h, X)
-    compose!(G, h, g, h)
-    return h
+    if Base.mightalias(g, h)
```
kellertuer, line 431: I was not aware of this function, that is really neat!

While it might resolve the issue here, I think this would even be nicer to not add it here but in every `compose!` implementation.

src/interface.jl
kellertuer: Oh I wrote my comment just in parallel to your question :)

src/interface.jl
Affie: @mateuszbaran suggested the function (`mightalias`) previously. 

`compose!` might also have aliasing problems, however the `exp!` here overwrites the value of `h` and therefore `g`.

Here is an example to further clarify:
```julia
julia> M = TranslationGroup(2);

julia> g = [2., 3];

julia> h = g;

julia> exp!(M, h, X)
2-element Vector{Float64}:
 0.1
 0.0
```
`g` is now set to the wrong value of `[0.1, 0.0]`

src/interface.jl
Affie: no problem, lets mark this as resolved an carry on in the other comment.

src/interface.jl
Affie: I do not think I am the correct person to try and fix the `HeisenbergGroup` (that is the only group where `exp!` still fails)

kellertuer (approved):
to me, this looks good. We can maybe collect a few things before we register this within a new version?

Unless this is urgent of course.

Affie:
> to me, this looks good. We can maybe collect a few things before we register this within a new version?

This is fine with me. I will use the branches and continue working through the update to LieGroups to make sure nothing else is missing/broken.
Perhaps the SGal PR should be included in the next release because it also has 2 bug fixes.

kellertuer:
One idea is, that I merge this and you can just pull main again locally and work on that :) Maybe that is a bit easier to have in mind also for other fixes.

Affie:
> One idea is, that I merge this and you can just pull main again locally and work on that :) Maybe that is a bit easier to have in mind also for other fixes

Yes merge please. I’ll merge main into the SGal branch and finish that. It looks the SGal branch is the last that I require for RoME tests to pass again.

Issue #52, opened 2025-06-24
Missing stuff from Manifolds.jl


mateuszbaran:
Stuff that still needs to be moved before removing groups from Manifolds.jl:

- [ ] `optimal_alignment`
- [x] `ColumnwiseMultiplicationAction`
- [x] `ComplexPlanarRotation`
- [x] `QuaternionRotation`
- [x] `quaternion_rotation_matrix`
- [x] `TranslationAction`
- [x] `RotationTranslationAction`
- [x] `RotationTranslationActionOnVector`
- [x] `ColumnwiseSpecialEuclideanAction`
- [x] `RotationAroundAxisAction`
- [ ] `AbstractCartanSchoutenConnection`, its subtypes and all relevant methods
- [x] `jacobian_exp_inv_argument(M::SpecialEuclidean{TypeParameter{Tuple{2}}}, p, X)` and `jacobian_exp_inv_argument(M::SpecialEuclidean{TypeParameter{Tuple{3}}}, p, X)` (probably as methods of `jacobian_exp_argument`?)
- [x] mention `adjoint_matrix` in transition; check if all methods have been moved
- [x] suggest replacements for `SpecialEuclideanInGeneralLinear`'s `embed` and `project` in the transition page.
- [ ] `get_orbit_action`

I'm planning on keeping the list dynamic. Stuff not worth moving will be crossed out or not mentioned here. I haven't looked through Manifolds.jl particularly thoroughly so I might have missed something relevant.

kellertuer:
Thanks for checking and doing a list.

I had hoped I was careful enough, maybe not – for the actions I thought, we are more general now here anyways, sine. the action here – if I recall correctly – has both `G and M` (the group and the manifold that acts on) stored internally so the separate types should hopefully be obsolete.

mateuszbaran:
If they are obsolete, then specifying a replacement in transition should be enough.

kellertuer:
I thoght about all the `Action`s during JuliaCon and it is unclear to me why and where we need all those.
For example the `RotationTranslationAction` is nothing else then using the group operation basically; so for all those types I am unsure what to do because for me there is nothing to implement – and then those would be useless structs/types.

mateuszbaran:
[...]

PR #60, opened 2025-07-31
Porting the SpecialGalileanGroup from RoME

Affie:
To close #11 
This is still a work in progress. 
- [x] fix `compose` and `inverse`, likely an issue with the semidirect `GroupAction`.
- [x] Remove temporary included `RecursiveArrayTools: ArrayPartition` 
- [x] Look at a replacement for `Rotations: RotationVec` and maybe remove.
- [x] Add documentation

PR #65, opened 2025-08-15
Transfer more elements from Manifolds.jl

kellertuer:
This adresses a few points from #52 

* [x] `adjoint_matrix` is now mentioned in the docs
* [x] the `apply` and `diff_apply` now correctly allocate a point on the manifold and a tangent vector, respectively
* [x] `jacobian_exp` (formerly `jacobian_exp_argument`)
* [x] add the jacobian also to the transition page once it is finished.
* [x] maybe add a note on left/right/actual(sorry, math precision here) Jacobians either in the transition or the interface page? See  [here](https://juliamanifolds.github.io/Manifolds.jl/stable/misc/notation/#Comparison-with-notation-commonly-used-in-robotics) for some notes we could take over.
* unify `<:AbstractGroupActions` from `Manifolds.jl` to `AbstractGroupActionTypes` there are
  * [x]  `ComplexPlanarRotation` is now a `LeftMultiplicationGroupAction` with corresponding Lie group and manifold.
  * [x]  `GroupOperationAction` is now e.g. https://juliamanifolds.github.io/LieGroups.jl/stable/interface/actions/#LieGroups.LeftGroupOperationAction or its right cousin.
  * [x]  `Manifolds.ColumnwiseMultiplicationAction` should be a `ColumnwiseGroupAction{LeftMultiplicationGroupAction}` of e.g. SO(n) and a corresponding manifold that has columns (like a matrix manifold)
  * [x]  `Manifolds.ColumnwiseSpecialEuclideanAction` is a be `ColumnwiseGroupAction{LeftMultiplicationGroupAction}` of e.g. SE(n) and a corresponding manifold that has columns (like a matrix manifold)
  * [x]  `Manifolds.RotationAroundAxisAction` should get its own type to store an axis
  * [x]  `Manifolds.RowwiseMultiplicationAction` should be `LeftMultiplicationGroupAction` of e.g. SO(n) and a corresponding manifold that has rows (like a matrix manifold) or better a `RightMultiplicationAction`?
  * [x]  `QuaternionRotation` is now a `LeftMultiplicationGroupAction` with corresponding Lie group (like SO(n)) and manifold (usually some Rn).
  * [x]  `RotationAction` is now a `LeftMultiplicationGroupAction` with corresponding Lie group (SO(n) and manifold (usually some Rn).
  * [x]  `RotationTranslationAction` should be checked but possible as a `LeftMultiplicationGroupAction` with corresponding Lie group (SE(n) and manifold (usually some Rn?).
  * [x] `RotationTranslationActionOnVector` is now a `LeftMultiplicationGroupAction` for SE(n) and Rn.
  * [x]  `TranslationAction` should get its own type since compared to the left_mult this is abelian, maybe `AdditionGroupAction`?

For the `jacobian_exp_argument` last one I got a bit confused. 
* there is [the ManifoldsDiff interface](https://juliamanifolds.github.io/ManifoldDiff.jl/stable/library/#ManifoldDiff.jacobian_exp_argument) which has a basis for both domain and codomain (ok, both are still tangent spaces but for completeness, fine...)
* and exactly one using that, namely on [rotations](https://github.com/JuliaManifolds/Manifolds.jl/blob/db305e085ae5608310c10a6d19e3524a118b6e81/src/manifolds/Rotations.jl#L207-L213) but it has neither of the bases – so magically works on _some_ basis?
* and the old function name was only used for one Lie group, namely [SE(n)](https://juliamanifolds.github.io/Manifolds.jl/stable/manifolds/group/#Manifolds.jacobian_exp_inv_argument-Tuple{GroupManifold{ℝ,%20ProductManifold{ℝ,%20Tuple{TranslationGroup{ManifoldsBase.TypeParameter{Tuple{2}},%20ℝ},%20SpecialOrthogonal{ManifoldsBase.TypeParameter{Tuple{2}}}}},%20Manifolds.SemidirectProductOperation{RotationAction{LeftAction,%20TranslationGroup{ManifoldsBase.TypeParameter{Tuple{2}},%20ℝ},%20SpecialOrthogonal{ManifoldsBase.TypeParameter{Tuple{2}}}}}},%20Any,%20Any}) but that also does not specify any basis

1. Which bases are used in points 2 and 3?
2. Would Rotations also be the one that is “invariant” so be the one on SO(n) (only n=2,3 implemented)
3. Is there a good test we could implement in the test suite for this new function?
4. And also, we would then extend the `ManifoldDiff` interface here?

For there I would need a bit of help @mateuszbaran 
I also have found out, why I did not add this function here in the first place. I did not forget it. It was introduced after I started and already had a list I was working on, see https://github.com/JuliaManifolds/Manifolds.jl/pull/767


mateuszbaran:
Thanks for working on this :+1: 

> * and exactly one using that, namely on [rotations](https://github.com/JuliaManifolds/Manifolds.jl/blob/db305e085ae5608310c10a6d19e3524a118b6e81/src/manifolds/Rotations.jl#L207-L213) but it has neither of the bases – so magically works on _some_ basis?

If there are no bases, it's assumed `DefaultOrthonormalBasis()` is used. If there are no basis arguments it means no other bases are supported by a method. It should still accept explicitly passing those bases but this feature was probably forgotten.

> * and the old function name was only used for one Lie group, namely [SE(n)](https://juliamanifolds.github.io/Manifolds.jl/stable/manifolds/group/#Manifolds.jacobian_exp_inv_argument-Tuple%7BGroupManifold%7B%E2%84%9D,%20ProductManifold%7B%E2%84%9D,%20Tuple%7BTranslationGroup%7BManifoldsBase.TypeParameter%7BTuple%7B2%7D%7D,%20%E2%84%9D%7D,%20SpecialOrthogonal%7BManifoldsBase.TypeParameter%7BTuple%7B2%7D%7D%7D%7D%7D,%20Manifolds.SemidirectProductOperation%7BRotationAction%7BLeftAction,%20TranslationGroup%7BManifoldsBase.TypeParameter%7BTuple%7B2%7D%7D,%20%E2%84%9D%7D,%20SpecialOrthogonal%7BManifoldsBase.TypeParameter%7BTuple%7B2%7D%7D%7D%7D%7D%7D,%20Any,%20Any%7D) but that also does not specify any basis

Yes, because default ONB was used as well and no other basis is supported. It wasn't used anywhere else because it's a niche functionality that is fairly technical to derive so no one bothered.

> 3\. Is there a good test we could implement in the test suite for this new function?

I think it would be good to test if `jacobian_exp_argument` and `diff_exp_argument` on basis vectors give results that agree with each other (see the default implementation of `jacobian_exp_argument!` in ManifoldDiff.jl)

> 4\. And also, we would then extend the `ManifoldDiff` interface here?

Yes, I think that's necessary.

kellertuer:
> Thanks for working on this 👍

Thanks, the answers already help :)

> > * and exactly one using that, namely on [rotations](https://github.com/JuliaManifolds/Manifolds.jl/blob/db305e085ae5608310c10a6d19e3524a118b6e81/src/manifolds/Rotations.jl#L207-L213) but it has neither of the bases – so magically works on _some_ basis?
> 
> If there are no bases, it's assumed `DefaultOrthonormalBasis()` is used. If there are no basis arguments it means no other bases are supported by a method. It should still accept explicitly passing those bases but this feature was probably forgotten.

Ah, that was not clear to me and is also a bit hard to see in the implementation. The problem is then, that in LieGroups the default is the VeeHat one – and I am also not sure, whether the given one is on the LieAlgebra or the tangent bundle. With the representation of Rotations it is probably the LieAlgebra?
Then I mainly have to check how much DefaultONB and VeeHat are different (maybe here they are not).

> > * and the old function name was only used for one Lie group, namely [SE(n)](https://juliamanifolds.github.io/Manifolds.jl/stable/manifolds/group/#Manifolds.jacobian_exp_inv_argument-Tuple%7BGroupManifold%7B%E2%84%9D,%20ProductManifold%7B%E2%84%9D,%20Tuple%7BTranslationGroup%7BManifoldsBase.TypeParameter%7BTuple%7B2%7D%7D,%20%E2%84%9D%7D,%20SpecialOrthogonal%7BManifoldsBase.TypeParameter%7BTuple%7B2%7D%7D%7D%7D%7D,%20Manifolds.SemidirectProductOperation%7BRotationAction%7BLeftAction,%20TranslationGroup%7BManifoldsBase.TypeParameter%7BTuple%7B2%7D%7D,%20%E2%84%9D%7D,%20SpecialOrthogonal%7BManifoldsBase.TypeParameter%7BTuple%7B2%7D%7D%7D%7D%7D%7D,%20Any,%20Any%7D) but that also does not specify any basis
> 
> Yes, because default ONB was used as well and no other basis is supported. It wasn't used anywhere else because it's a niche functionality that is fairly technical to derive so no one bothered.

We should still at least document that carefully and forward the default basis (if they are passed) to this case then.

> > 3. Is there a good test we could implement in the test suite for this new function?
> 
> I think it would be good to test if `jacobian_exp_argument` and `diff_exp_argument` on basis vectors give results that agree with each other (see the default implementation of `jacobian_exp_argument!` in ManifoldDiff.jl)

That sounds like a very good idea, will work on that then.
 
> > 4. And also, we would then extend the `ManifoldDiff` interface here?
> 
> Yes, I think that's necessary.

Nice, the old implementation did that not that thoroughly, I think.

mateuszbaran:
> Ah, that was not clear to me and is also a bit hard to see in the implementation. The problem is then, that in LieGroups the default is the VeeHat one – and I am also not sure, whether the given one is on the LieAlgebra or the tangent bundle. With the representation of Rotations it is probably the LieAlgebra?
> Then I mainly have to check how much DefaultONB and VeeHat are different (maybe here they are not).

IIRC they differ by some sqrt(2) factors here and there. So the formulas will be different. I guess to avoid confusion LieGroups.jl could require the basis to be explicitly passed and not implement the basis-free variant.

kellertuer:
Over in #67 I noticed that I was very sloppy with my action types and you are correct we need more than just the GroupOperationAction. That currently makes my nice new defaults fail, since it tries to apply the group operation (of matrices) to a matrix vector multiplication. 

So either we should first finish this or tweak the PR #67 to at least use an action that does not default to use compose to perform the action on SE(n), since that fails with a result of wrong dimensions.
I think my original idea was that `GroupOperationAction` could have a loose interpretation as a function $f_g(x) = g \circ x$ like the group operation but also on different domains for x (mapping then to y from the same space (this is a left action of course).  I now see that this was not very clever to take that shortcut, maybe.
Thought the action does know its domain and range so it should be fine. Not sure it actually is by now.

However maybe an action type like `MatrixVectorMultiplcationAction` was also missing before? At least for now I am a bit stuck, whether to continue here or there.

kellertuer:
I got unstuck today 🤓 I also have an alternative already for the rotationAction and will document that over on the other PR.

mateuszbaran:
I see; `LeftMultiplicationGroupAction` sounds good and would cover some of the missing cases. A loose interpretation of `GroupOperationAction` could indeed be problematic.

kellertuer:
Nice! Then I mainly have to find one of the left/righty things that currently make the tests fail. I hopefully find a bit of time the next days, when I have settled in in Erlangen.

And I fully agree that the previous code misused the GruptOpAction, but luckily that is fixed now.

kellertuer:
I updated the list above to list all existing (old, Manifolds.jl) group actions and might try to transfer them to types next.

I also tried to read up ton left/right Jacobians. I think they are maps $T_gG \to \frak g$ or `\frakg g \to T_gG$ (as opposed to our $\frak g\to\frak g$) but from https://arxiv.org/pdf/1812.01537 it is impossible to find that our, since even their definition of their $\oplus/\ominus$ is vague and unclear to me.

mateuszbaran:
> * [ ]   `Manifolds.RowwiseMultiplicationAction` should be `LeftMultiplicationGroupAction` of e.g. SO(n) and a corresponding manifold that has rows (like a matrix manifold) or better a `RightMultiplicationAction`?

`RightMultiplicationAction` might again cause confusion: Is it a right-action or action multiplying from the right? Does this name even correspond to what `apply` is doing? It's supposed to be implemented like this:
```
function apply(::LeftRowwiseMultiplicationAction, a, p)
    return (a * p')'
end
```
and if there is no error it should behave like a left action. I think `RowwiseMultiplicationAction` might be the least confusing name here.

> I also tried to read up ton left/right Jacobians.

I think the definition from our manuscript should be clear:
* The right Jacobian of $f$ at $p$ is the standard Jacobian of $f$ at $p$.
*  The left Jacobian of $f$ at $p$ is the standard Jacobian of $q \mapsto f(q \circ p) \circ f(p)^{-1}$ evaluated at $q=e_{\mathcal{M}_1}$.

kellertuer:
sure `RowiseMultiplicationAction` and then also `ColumnwiseMultiplicationAction` sounds fair enough.

For the lefties and righties again, now for Jacobians:

* so the right jac is the differential from T_pG to T_f(p)G in coordinates?
* and the left jac is the differential from TeG to TeG in coordinates (starting in q=e yields the first TeG and the f(p)^-1 is a pullback from T_f(p)G to TeG)?

that would be nice to document, our default is then the left one. Literature for that would of course be great.
(I thought for a while they where the mixed variants that would also be possible)

kellertuer:
ah or one could do a `ColumnwiseActionType{A}` where `A` is another action type :) and the same for rows.

mateuszbaran:
> For the lefties and righties again, now for Jacobians:
> 
> * so the right jac is the differential from T_pG to T_f(p)G in coordinates?
> * and the left jac is the differential from TeG to TeG in coordinates (starting in q=e yields the first TeG and the f(p)^-1 is a pullback from T_f(p)G to TeG)?

Yes, that sounds right.

> that would be nice to document, our default is then the left one. Literature for that would of course be great.
> (I thought for a while they where the mixed variants that would also be possible)

Yes, the left one should be the default due to our tangent vector representation. I'm not sure what's the best literature reference for that is though. Micro Lie is OK-ish though they use a weird notation.
Mixed variants are of course also possible:
* standard Jacobian of $q \mapsto f(q \circ p)$ evaluated at $q=e_{\mathcal{M}_1}$.
* standard Jacobian of $q \mapsto f(q) \circ f(p)^{-1}$ evaluated at $q=p$.

I think they are less useful though? And I'm not even sure if there is a standard naming for them.

kellertuer:
NIce!

No, the mixed ones were just my confusion then, let's not add them.

But then I have a plan how to continue here, besides finding a reference for the left/right Jacobian reference (in the transition I want to mention it).

Affie:
Other than Sola, Chirikjian vol 2 also has left and right Jacobians on p26. 

I understood that LieGroups.jl are using right Jacobians (right as Sola definition, think Chirikjian also has the same right).

I do not understand the definition in your manuscript:
* The right Jacobian of $f$ at $p$ is the standard Jacobian of $f$ at $p$.
*  The left Jacobian of $f$ at $p$ is the standard Jacobian of $q \mapsto f(q \circ p) \circ f(p)^{-1}$ evaluated at $q=e_{\mathcal{M}_1}$.

I also don't understand Chirikjian, but thought I at least now understood LieGroups.jl, but it seems I have it the wrong way around?

> since even their definition of their ⊕/⊖ is vague and unclear to me.

LieGroups.jl uses right ⊕ and ⊖ eq 25 and 26. To help you make the connection with LieGroups.jl:
- g ⊕ X = exp(G, g, X)
- g ⊖ h = log(G, g, h)

kellertuer:
I will never define those plus/minusses, I find them super misleading and imprecise. Also with you mentioning which ones we use, I do not see where we use them nor how nor why. But the plus is also not commutative, so I feel that is very dangerous.

I can check Chirikjian, but for the ones now documented, those are the two you claim you do not understand.

And no, LieGroups.jl is not using right Jacobians, it is using left ones.

So the first is really just the one on manifolds. If you have $f: \mathcal M \to \mathcal M$, then $Df(p)$ “takes” a change at the point in the domain $p$, so from $T_p\mathcal M$ and “returns” how such a change would change the output, so that is a change (tangent vector) in $T_{f(p)}\mathcal M$.
This is the usual differential in `Manifolds.jl`. For a matrix representation we use a basis of the first and the second tangent space. Then we obtain a right Jacobian.

The second one has the advantage that it takes into account how we usually represent tangents anyways, namely in the Lie algebra. In `LieGroups.jl` we denote the second one by $df(p)$ and due to the q upfront it takes as input vectors from the Lie algebra (note that the classical differential we evaluate at e) and due to the pullback $from(p)^{-1}$ the result is again in the Lie algebra.
If we now take (just) one basis of the Lie algebra - we obtain a matrix representation of $df(p)$ – the left Jacobian. That is what we use.

By the way this difference is also what killed my `diff_compose` on semidirects: I computed the first differential, because I know a bit better how to compute them, but in LieGroups.jl we would need the second one. And those are still tough for me to compute.

mateuszbaran:
> I understood that LieGroups.jl are using right Jacobians (right as Sola definition, think Chirikjian also has the same right).
> 
> I do not understand the definition in your manuscript:

This is quite technical unfortunately. To be precise, for a Jacobian you need a function $f \colon \mathbb{R}^n \to \mathbb{R}^m$. You can then calculate the Jacobian at a point $x \in \mathbb{R}^n$ and get a standard $m \times n$ matrix. If a function is not defined between Euclidean spaces, we also need charts to make it work. If you want, I can explain it in a meeting -- it will be easier than guessing which part of the definition is unclear. The result only depends on the differentials of charts at the point of evaluation and the value, which defines bases, so that's why we can speak about coordinates of a Jacobian in bases (at $p$ and $f(p)$).

Micro Lie keeps mixing Lie algebra and coordinates in a basis so it's quite hard to follow their definition of Jacobians.

I fully support not using pluses and minuses here. They cause confusion and don't simplify anything much.

> LieGroups.jl uses right ⊕ and ⊖ eq 25 and 26. To help you make the connection with LieGroups.jl:

You probably forgot `vee` and `hat` somewhere but it's hard to tell because Micro Lie theory is inconsistent about taking coordinates. Also, LieGroups.jl doing computations in the Lie algebra also means that `compose` and its differential is needed in some variants.

kellertuer:
@mateuszbaran does this look correct? 

https://github.com/JuliaManifolds/LieGroups.jl/blob/e0c5a117741ddcf78c5cea7e10453151b24e7841/ext/LieGroupsRecursiveArrayToolsExt/special_euclidean_group_RAT_ext.jl#L7-L50

I did take over the code, I tried to extend the doc string to the correct form. If this is ok, one could also introduce a right multiplication for the other case you did, though there the original docs are a bit too vague for me to understand what it is meant to do, see https://github.com/JuliaManifolds/Manifolds.jl/blob/5f309f750c0538685a8eaa2c606c5fa8235c4127/ext/ManifoldsRecursiveArrayToolsExt/rotation_translation_action_rat.jl#L9-L16 – that is a bit too short for my small little mind to follow where that is from and what it is for.

Affie:
On ⊕ and ⊖, yes there is a hat/vee `c = vee(LieAlgebra(G), X)` to match the paper exactly. I also do not like ⊕ and ⊖ (⊖ especially since it is ambiguous) and was just trying to help make the link between the paper and LieGroups.jl notation (with julia sorting out the dispatch :smiley: ).
Here it is again as I understand it but using typical LieGroups.jl symbols:
- h = g ⊕ c <--> h = exp(G, g, X)
- c = g ⊖ h <--> X = log(G, g, h)

Affie:
I think your explanations helped a bit, I think it's mostly the notation I do not follow exactly; it is now the 4th way the same thing is written, Sola, LieGroups.jl, Chirikjian all use different notations. 
If I look at the actual code in LieGroups.jl it matches with the right Jacobians in Sola. 
Trying to match your notation, this is what I now think....
1. The standard Jacobian of $f$ at $p$ used in Manifolds.jl
2. The left Jacobian: ... $q \mapsto f(q \circ p) \circ f(p)^{-1}$...
3. The right Jacobian used in LieGroups.jl: ... $q \mapsto f(p)^{-1} \circ f(p \circ q)$ ...

But it seems that if there is a left and right somewhere I will mix things up.

kellertuer:
I think there is even a mixup about the term “Jacobian”. If we look at Chirikjan, p. 26 - his “Jacobian” is basically a push forward or pullback of a differential. (q to g), where I think he refers to the tilde g he defined before and then said he drops the tilde since it is clear from context to distinguish g and tilde g – and the line after I am not sure which one he means, since it does not have a (t) or a (q) any longer. Maybe it is tilde g because otherwise it would not make so much sense.
If I understand them correctly the in the end do the “change of viewpoint” from tangent spaces to Lie algebra in some sense but he never motivates why he introduces them. But it does seem to be a “quantity of interest” (why ever)

Those Jacobians are not the Jacobians we talk about here. 

As Mateusz already wrote, Jacobian in general (see https://en.wikipedia.org/wiki/Jacobian_matrix_and_determinant) generalises the differential – while I would even call the “derivative information” of the multivariate $f$ from the article a differential, namely the operator $Df(p)[X]$ that provides the information how f would (infinitesimally) change at p when moving in direction X. Since that is a linear operator it has a matrix representation. _That_ matrix is the Jacobian (in Eucidliean terms) and implicitly it uses the canonical ONB of R^n both for X and for the resulting vectors.

Yes 1. is the classical differential – and with bases of the tangent spaces (at p and f(p)) we would get the Jacobian. We could say “Riemannian Jacobian” just to give it a name other than just Jacobian.
The second is actually the one we use in LieGroups.jl

The third one I have not yet seen, as Mateusz wrote, the first is usually called the right Jacobian? But that is why I would like to have a better literature about this, that talks about differentials and Jacobians (in the sense we talk about here) on Lie groups, because your third case does look interesting :)

kellertuer:
And just half a tea later:

Your third case would be the correct thing to do if we identify the tangent spaces with the Lie algebra via right-invariant vector fields – but the default (and only available thing for now) are left invariant ones (which yields the second case).

Overall, I now mention left/right Jacobians in the notation only exactly once – and the reason is that the term is misleading by itself.

I would call case 1 either the Riemannian Jacobian or the Jacobian (maybe add w.r.t. tangent spaces)
And I would call 2/3 the Jacobian (w.r.t. the Lie algebra) – where one could add which invariance we used to identify tangent spaces.

mateuszbaran:
> @mateuszbaran does this look correct?

Yes, this looks right.

> https://github.com/JuliaManifolds/Manifolds.jl/blob/5f309f750c0538685a8eaa2c606c5fa8235c4127/ext/ManifoldsRecursiveArrayToolsExt/rotation_translation_action_rat.jl#L9-L16 – that is a bit too short for my small little mind to follow where that is from and what it is for.

IIRC that right action was only added for completeness and might not be particularly useful. I think we can skip it. Basically, this right action of `a` on `p` is the same as the left action of `inv(G, a)` on `p`.

> If I look at the actual code in LieGroups.jl it matches with the right Jacobians in Sola.
> Trying to match your notation, this is what I now think....

I honestly don't remember which one was used there but it might have been the right one. Ultimately we always have to check against finite differences and multiply by differential of conjugation (the adjoint matrix) as needed. But it would definitely be good to be consistent here about it. Implicit translation to and from identity does seem to introduce some confusion here.

kellertuer:
Cool, then let's only add the right action if someone actually needs it – and yeah it also looks like the left-inv action (which is a right action, I know ;))

kellertuer:
For completeness, I started a draft on the notations page about Jacobians, see https://juliamanifolds.github.io/LieGroups.jl/previews/PR65/notation/#About-differentials-and-Jacobians (and yes for now I messed up the math/code environments) – I would like to make this understandable and precise and mention we would not write left/right throughout LieGroups.jl – and hopefully also provide a reference.

kellertuer:
I fixed a few typos in the doc string linked above and finished documenting and testing `apply` and `diff_apply` for SE(n) acting on Rn – in all 3 variants.
For `diff_group_apply` the old code did not provide documentation and only covered the identity so I left that out.
1 action down, 8 to got.

For the Left/Right Jacobian – Johannes might be right with the third case (right Jacobian for right invariant vector fields) – then the first is not called right one – but really more like tangent space one. It would still be great to have literature on this.

_edit:_ I think that left/right Jacobians in the sense of Micro-Lie are more like pushforwards/pullbacks and not related to the Jacobian we consider here. Maybe we should omit mentioning left/right Jacobians at all? Then one could add a note that for right.invariant vector fields the Lie algebra diff would change (and how) and leave it at that concerning Jacobians.

If we ever come to the need to define functions for the (actual) things called left/right Jacobians, we should check how to best call them, but they do not involve functions f but curves, so they are to some extend different things.

Affie:
I think part of the left/right confusion from Chirikjian is explained in footnote 15:
"_The “l” and “r” convention used here for Jacobians and for vector ﬁelds is opposite that
used in the mathematics literature. The reason for the choice made here is to emphasize the
location of the “the most informative part” of the expression. In Jacobians, this is the location
of the partial derivatives. In vector ﬁelds, this is where the components deﬁning the ﬁeld appear._"

So, the left invariant vector fields in other literature may be right invariant in Chirikjian?

> I think that left/right Jacobians in the sense of Micro-Lie are more like pushforwards/pullbacks and not related to the Jacobian we consider here

I didn't know there were different Jacobian, just different definitions of tangent vectors and though we were looking at push forward as per https://en.wikipedia.org/wiki/Pushforward_(differential)

kellertuer:
Nirgs, I start to not like Chirikjan. That reads like a strange note.
_edit:_ it reads a bit like: The left hand? Lets call it right hand, because it has the thumb on the right and let's call the right hand left hand, because it has the thumb on the left! The thumb is where catching something really takes place and that is what we focus on! 😭

My main problem is, that in what _we_ do here we have a function f, look at its differential and then the differentials matrix representation -> Jacobian.

Chirikjan does not have such f. So what is he differentiating then?

Affie:
I think in Chirikjian it is the "Left Jacobian"  (Jₗ) and "Right Jacobian" (Jᵣ) specifically. They are the left and right Jacobians of the exponential function. I gave the reference as that is one of the only sources where I know they define left and right when it comes to Jacobians. From Sola: "The right Jacobian maps variations of the argument τ into variations in the local tangent space at Exp(τ)"

Affie:
In a paper by Mohony https://arxiv.org/pdf/2403.16411 he defines left/right Jacobians as you suggest.  "The left trivialised Jacobian is referred to as the right Jacobian in much of the key literature" - referring to Chirikjian.

kellertuer:
Ah, that one I missed, since I was looking for functions mapping from G to G - but sure with the trivial identification of T_g(TgG) = T_gG, exp is also included, for sure.

For the linked paper...if I now only knew what left trivialisation refers to...have to read up on that first.
_edit:_ ah it is only again a language barrier. What they mean is the identification of the tangent spaces via left invariant vector fields (or the differential of the compose-left)
But Mahony and co-authors also note that Chirikjan did switch left/right there. Really a strange choice. I will now also start to refer to my left hand as the right hand for similar reasons.

Affie:
Does that clear it up .... LieGroups.jl uses left invariant vector fields and the Jacobians corresponds to the right Jacobians in Sola and Chirikjian? 

My left hand is now my right hand if my palm is facing down but if my palm is facing up then it is my left hand again?

kellertuer:
I have not yet understood Sola. Do they use the same naming scheme as Chirikjian?

I think overall I would still avoid the term left/right jacobian, now even more that to be precise it is the right/left invariant identification used on the default Jacobian of (just) the exponential map.

If you let your palm face up on your (now) right hand, sure the thumbnail is facing left again; but I would prefer the term inverse-right hand then.

kellertuer:
I generalised Row and Column actions which turned out to be very nice and generic; I documented what was changed – though not yet sure how to check those.

The remaining four ones should be special cases of existing types (`LeftMultiplicationGroupAction` with corresponding group and manifold), should however be still tested and documented in the `transition.md`

Affie:
Hi, I tried to write down how I matched LieGroups.jl with Sola and your paper (with a bit of AI help for the notation, so hope i got it right):

### 3. Identification of the tangent spaces via left invariant vector fields, Left-trivialized
 (Represent tangent vectors as gX in a matrix Lie group; Sola right Jacobian, ie. using right ⊕ and ⊖):

```math
λ(h) := f(g)^{-1} f(g ∘ h) = L_{f(g)^{-1}}( f( L_g(h) ) )
```

As a composition:

```math
λ = L_{f(g)^{-1}} ∘ f ∘ L_g
```

Differentiating at the identity $e$:

```math
dλ|_e = d(L_{f(g)^{-1}})_{f(g)} ∘ d f_g ∘ d(L_g)_e
```

Example `diff_inv` with $f(g)=g^{-1}$:
1. Start with $X ∈ 𝔤$.  
2. $dL_g|_e[X] = g X$.  
3. $d f_g[gX] = - g^{-1} (gX) g^{-1} = - X g^{-1}$ (using the matrix differential for the inverse $d f_g[X] = - g^{-1} X g^{-1}$)  
4. Transport back with $dL_{f(g)^{-1}}$; ($f(g) = g^{-1}$ so $f(g)^{-1} = g$):  
   $dL_g[- X g^{-1}] = g(- X g^{-1}) = - g X g^{-1} = - Ad_g[X]$.

 
$-Ad_g$ matches `diff_inv` and  Sola eq 62

kellertuer:
Besides the total mixup Sola seems to do on left and right, that looks ok.
I have no clue how gX – the differential of left multiplication can lead to right plus/minus/jacobians. That has zero logic for me.

I think I would prefer to call this the Jacobian with respect to the Lie algebra representation using left-invariant vector fields.
One could use tangent space representation instead – or for the Lie algebra representation also right-invariant vectorfields.
Tangent space would be used in most cases on Manifolds.jl (for example not for the RotationMatrices or the Unitary matrices, but in nearly all other cases).
Right-invariant vector fields are currently not used anywhere, but if there is a large need one could parametrise the Lie algebra or something like that.

kellertuer:
I just transferred the last action.

Now just a bit of code coverage for that action and a function for the jacobian in the test suite (and calling that for one Lie group) is missing. So this PR is nearly finished.

kellertuer:
I ran into a bit of fuzz with `jacobian_exp_argument` and its differences to the one in `ManifoldDiff.jl`.
Then I noticed the actual Jacobian implemented was the one of the Lie group exponential function, which does not even have a base point, but just an argument. So now this PR introduces `jacobian_exp` instead.

Now that I am writing this, it might still have the `g` argument too much then, but I think it fits better to the interface design here with the new name (just that I might have to remove the g to be more persistent then).

But with this, the PR should be complete - let me know what you think! If all is fine but the removal of g, I can do that as a final step.

mateuszbaran:
> Then I noticed the actual Jacobian implemented was the one of the Lie group exponential function, which does not even have a base point, but just an argument. So now this PR introduces `jacobian_exp` instead.

Maybe let's just have Jacobian of the `exp` associated with the connection, not the Lie group exponential? That way we don't need to introduce new functions.

> But with this, the PR should be complete - let me know what you think! If all is fine but the removal of g, I can do that as a final step.

Cool! I will check this soon.

kellertuer:
> > Then I noticed the actual Jacobian implemented was the one of the Lie group exponential function, which does not even have a base point, but just an argument. So now this PR introduces `jacobian_exp` instead.
> 
> Maybe let's just have Jacobian of the `exp` associated with the connection, not the Lie group exponential? That way we don't need to introduce new functions.
> 

I do not understand what you mean here. The previous implementation was called argument, had even the `g` in there and ignored it. So it was misleading, I think.

mateuszbaran:
> I do not understand what you mean here. The previous implementation was called argument, had even the `g` in there and ignored it. So it was misleading, I think.

Was it incorrect? If the differential or Jacobian does not depend on `g`, it can be ignored.

kellertuer:
I am not so sure. I think it was really just the one of the exponential function, not the exponential map. But maybe we otherwise leave this out here and someone else with more knowledge can do that in a next PR.
I did what I could with my existing knowledge. but that is probably then (yet again) not sufficient.

So maybe lets then not do this on this PR maybe.

mateuszbaran:
OK, then let's tackle that in a new PR.

docs/src/tutorials/transition.md
```diff
@@ -17,11 +17,14 @@ Lie algebra elements (vectors) always `X, Y`.
...
 
-The list is alphabetical, but first lists types, then functions
+The list lists first types, then functions. Within both blocks, the order is alphabetical
 
 | `Manifolds.jl` | `LieGroups.jl` | Comment |
 |:---------- |:---------- |:-------------- |
 | `AdditionOperation` | [`AdditionGroupOperation`](@ref) | |
+| `ColumnwiseMultiplicationAction` | [`ColumnwiseGroupAction`](@ref)`(`[`LeftMultiplicationGroupAction`](@ref)`)` | within a [`GroupAction`](@ref)`(action, group, manifold)` |
+| `ColumnwiseSpecialEuclideanAction` |  [`ColumnwiseGroupAction`](@ref)`(`[`LeftMultiplicationGroupAction`](@ref)`)` | within a [`GroupAction`](@ref)`(action, group, manifold)`, where the `group`is a [`SpecialEuclideanGroup`](@ref) |
+| `ColumnwiseSpecialEuclideanAction` | [`ColumnwiseGroupAction`](@ref)`(`[`LeftMultiplicationGroupAction`](@ref)`)` | within a within a [`GroupAction`](@ref)`(action, `[`SpecialEuclideanGroup`](@ref)`, `[`Euclidean`](@extref `Manifolds.Euclidean`)`(n)`. |
```
mateuszbaran: One entry for `ColumnwiseSpecialEuclideanAction` should be enough?

Affie:
I think `jacobian_exp_argument` does not depend on `g` on lie groups. As `diff_right_compose(G, g, h, X) = X` and `exp(G, g, X) = compose(G, g, exp(X))`

So, one can use the chain rule for the Jacobian/diff of `exp(G, g, X)` so I would guess you only need `jacobian_exp`?

docs/src/tutorials/transition.md
kellertuer: Oh, indeed ;) When doing to many PRs over too long time in parallel. I must have missed that.

kellertuer:
My main reason for the new function name is, that `exp(G,X)` does not need a distinction between argument and base point, like `exp(G,g,X)` does. And it might be that the differential with respect to the argument `X` is then just a chain rule - but that was not the case in the previous code. Also for the second we can also do a differential with respect to the base point `g` again.

mateuszbaran (approved):

kellertuer:
Oh, that was fast :) We could maybe release this together with the other PR #60 – and wait for it to bump version here?

PR #66, opened 2025-08-20
Improve how product and semidirect Lie group constructors act.

kellertuer:
For Product manifolds we currently have the following behaviour

```{julia-repl}
julia> using Manifolds

julia> M1 = Sphere(2); M2 = Euclidean(2); M3 = SymmetricPositiveDefinite(2)
SymmetricPositiveDefinite(2)

julia> M1 × M2 × M3
ProductManifold with 3 submanifolds:
 Sphere(2, ℝ)
 Euclidean(2; field=ℝ)
 SymmetricPositiveDefinite(2)

julia> ProductManifold(M1, M2, M3)
ProductManifold with 3 submanifolds:
 Sphere(2, ℝ)
 Euclidean(2; field=ℝ)
 SymmetricPositiveDefinite(2)

julia> ProductManifold(M1, ProductManifold(M2,M3))
ProductManifold with 2 submanifolds:
 Sphere(2, ℝ)
 ProductManifold(Euclidean(2; field=ℝ), SymmetricPositiveDefinite(2))
```
where the first also implies that both

```{julia-repl}
julia> M1 × ProductManifold(M2,M3)
ProductManifold with 3 submanifolds:
 Sphere(2, ℝ)
 Euclidean(2; field=ℝ)
 SymmetricPositiveDefinite(2)
julia> ProductManifold(M1, M2) × M3
ProductManifold with 3 submanifolds:
 Sphere(2, ℝ)
 Euclidean(2; field=ℝ)
 SymmetricPositiveDefinite(2)
```

or in other words: `×` always “splats” its involved Product manifolds, while the direct constructor does not. This allows to produce both results.

This PR aims to provide the same for the `ProductLieGroup` both to be consistent and to be able to generate both the direct product of multiple Lie groups as well as the “nested” variant.

* [x] `G1 × G2 × G3` works (already before the PR)
* [x]  `ProductLieGroup(G1, G2, G3)`should produce the same (before the PR it errored)
* [x] bsted cases like `ProductLieGroup(G1, ProductLieGroup(G2,G3))`(already before)
* [x] `ProductLieGroup(G1,G2) × G3` and `G1 × ProductLieGroup(G1,G2)` work as in the second block (`×` “splats”)
* [x] check interoperability with the semidirect one, that that not accidentally “splats” but keeps a product of exactly 2.
* [x] add tests for points 2 and 5 of this list.

For the last point, before this PR, it used `×` which as shown above is the wrong one internally, since it shuld not “splat” the two parts of the semidirect product but really just have a product of two.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+* Fixed an issue where internally the product manifold in a `SemidirectProductLieGroup` was accidentally splashed
+* make `×` and `ProductLieGroup` behave the same way as `×` and `ProductManifold` do
 * Fixed an issue, where `exp!(G, h, g, X)` would return a wrong result if the input `g`and the output `h`are aliased (#63).
 
 ## [0.1.3] 2025-08-04
--- a/src/groups/product_group.jl
+++ b/src/groups/product_group.jl
@@ -58,14 +58,23 @@ function LinearAlgebra.cross(P1::ProductGroupOperation, P2::ProductGroupOperatio
 end
 
 """
-    ProductLieGroup(G, H, ...)
+    ProductLieGroup(G1, G2, ..., Gn)
 
-Return the [`LieGroup`](@ref) of the product of Lie groups `G` and `H`.
+Return the [`LieGroup`](@ref) of the product of Lie groups `G1`, `G2`, up to `Gn`
+and all following Lie groups.
+This can be considered as a vector of Lie group, where the vector is always
+of the same length as the number of provided Lie Groups.
 
-Alternatively, the short hand `G × H` can be used.
+If none of the Lie groups are product Lie groups themselves,
+this is equivalent to `G1 × G2 × ... × Gn`.
+
+For an example illustrating the differences see [`x`](@ref LinearAlgebra.cross(::LieGroup...)`(::LieGroup...)`.
 """
-function ProductLieGroup(G::LieGroup, H::LieGroup)
-    return LieGroup(G.manifold × H.manifold, G.op × H.op)
+function ProductLieGroup(lie_groups::LieGroup...)
+    return LieGroup(
+        ProductManifold([L.manifold for L in lie_groups]...),
+        ProductGroupOperation([L.op for L in lie_groups]...)
+    )
 end
 
 function ManifoldsBase.submanifold_components(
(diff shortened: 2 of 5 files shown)


Affie:
Please also add the equivalent ProductLieGroup to what was possible in Manifolds, continuing the example from above:
```
Ms = [M1, M3, M3]
julia> ProductManifold(Ms...)
ProductManifold with 3 submanifolds:
 Sphere(2, ℝ)
 SymmetricPositiveDefinite(2)
 SymmetricPositiveDefinite(2)
``` 

see PR discussion https://github.com/JuliaManifolds/LieGroups.jl/pull/60#discussion_r2288102081

Affie:
Also see the tests I create to try and cover all the different options:

https://github.com/JuliaManifolds/LieGroups.jl/blob/4090133da146a1d82f971f65be54fef848a7c8d3/test/groups/test_semidirect_product_group.jl#L93-L160

kellertuer:
Yeah sure, that is exactly the main thing to add here :) 

... and make cross ( `×` ) use it. And yeah I will check and maybe steal a few of your tests.

kellertuer:
Ah reading your tests I noticed even one thing one could improve a bit:
For a single argument `ProductLieGroups(G)` could be `G` is it is already a product LieGroup itself.
If not, then we can still form vectors of length 1 ;)

Affie:
> Ah reading your tests I noticed even one thing one could improve a bit:
For a single argument ProductLieGroups(G) could be G is it is already a product LieGroup itself.
If not, then we can still form vectors of length 1 ;)

Are you referring to these 2 cases? 
https://github.com/JuliaManifolds/LieGroups.jl/blob/4090133da146a1d82f971f65be54fef848a7c8d3/src/groups/product_group.jl#L80-L85

Yea, it would be good to add to the tests.

kellertuer:
I think the very first case is not necessary in my code, since I have the general case `ProductLieGroup(lie_groups::LieGroup...)` which covers that but the second is probably nice.

kellertuer:
Also the docs now cover similarities and differences in detail (up to typos that I have to fix). 

https://juliamanifolds.github.io/LieGroups.jl/previews/PR66/groups/product_group/#LinearAlgebra.cross-Tuple{Vararg{LieGroup}}

kellertuer:
This should be good to go. Can you check it once more and approve it? Then I can merge it later today after I had dinner – you merge it into SGal – and we resolved at least those technicalities.

Affie (approved):
Ok, I see the difference is that you can create nested products as well. That is fine with me.
I tested a few things locally and it looks good to me, just one comment is that the tests do not cover _"Fixed an issue where internally the product manifold in a `SemidirectProductLieGroup` was accidentally splashed"_ to prevent it from breaking again in the future.

kellertuer:
Ah, I forgot that test and added it right now (just noticed, that at home I have not yet set up runic, so let's hope for the best ;))

PR #67, opened 2025-08-22
Refactor the semidirect product Lie group to cover all cases from GroupManifolds.

Affie:
This possible bugfix was split out from #60 as it was not clear if it's a misunderstanding or a real bugfix.

- Compose:
    - Compose for RightSemidirectProductGroupOperation and LeftSemidirectProductGroupOperation uses the same formula
    - The formula and implementation match with the compose docstrings
- Inverse:
    - The old formula does not match with that of compose, it was missing a `^{-1}`.
- Tests
    - The action on SO(2) ⋉ SO(2) may have been wrong making the test invalid. The new test ⋉ reduces to just × (not 100% confident on this)
    - Added (SO(2), Tr(2)) with the options of left-left, left-right, right-right, and right-left
    - Also added a Semidirect product Heisenberg group that I'm not 100% sure of 

Something else:
The formula for [LeftSemidirectProductGroupOperation](https://juliamanifolds.github.io/LieGroups.jl/stable/groups/semidirect_product_group/#LieGroups.LeftSemidirectProductGroupOperation)
 
image

may not match [compose](https://juliamanifolds.github.io/LieGroups.jl/stable/groups/semidirect_product_group/#Manifolds.compose-Union{Tuple{%F0%9D%94%BD},%20Tuple{LieGroup{%F0%9D%94%BD,%20LeftSemidirectProductGroupOperation,%20%3C:ManifoldsBase.ProductManifold},%20Any,%20Any}}%20where%20%F0%9D%94%BD)
image
** note on what component the action is applied


kellertuer:
The easy thing: The docs of compose are wrong, the ones from left seem correct (I nearly wrote right, but that is misleading ;))

For the first: Yes both should use the same formula, but with “switched places” or compared to the docs, the left one has (h,n), then all formulae for the right have (n,h).

The SO(2) semidirect test ist not wrong. Just a bit boring. But it does not boil down to a direct product. Let me try to sketch that:
In this case we have that the action is really and literally the group operation $\sigma_{g}(h) = gh$. But then

$(g_1, h_1) \circ (g_2, h_2) = (g_1g_2, g_2h_1h_2)$

where on the left all operations are the SO(2) group operation, that is just matrix multiplication. And we have a left operation so g_2 is _left_ of h_1 – and we have a leftsemidirect product, so the _left_ is the one with the “normal” group operation.
So is is _nearly_ the direct product, and overall a bit boring.

Finally for the inverse and the new test, I have to carefully check and just opened the book before I saw this PR.

Affie:
Extra TODO: Add tests that `apply!(A, k, g, h)` is safe against aliasing if `k === h`

kellertuer:
To answer the “something else” point here as well after our discussion in the chat and some tinkering and sorting in my head

Main point: We will now switch to the variant of Hilgert&Neeb because that is more common.

Long form: There are three left/rights to decide on

1. left or right semidirect product, $\ltimes$ vs $\rtimes$ or informally: is the group on the left and the action applied to the right of the tuple or is the action applied on the left and the “pure group” component on the right., cf. Hilgert&Neeb Def 9.22 first (right) vs third (left) equation
2. left or right action, that is, whether an action behaves like a group operation from the left or right when concatenating actions.
3. in the action application, whether to apply the action to the first or second element, i.e. $\sigma_{g_2}(h_1)h_2$ vs $h_1\sigma_{g_1}(h_2)$. For now the second case seems more common (though Hilgert&Neeb also have both). Maybe this was formerly even forward/backward? I am not sure. We could in principle support this, but it would need a side-argument or so. Let's first fix this to the right case.

I updated the operation doc strings already, will look at compose and inv tomorrow then

kellertuer:
Thanks for the help on the doc strings, those are for today under heavy rework, since I noticed yesterday, that one thing we discussed was formerly the direction/side/forward/backward game in GroupManifolds. So the old Code does support all 8 variants we ended up with yesterday.

I think I found a good way to do that as well, which I will use as a chance to also document all 8 group operations, but then on the separate `compose` doc strings - as well as all inverses. 
I hope that after documenting all those, the different variant that are possible become much clearer.
I also avoided the terms direction (which was left/right), side (which was left/right) and forward/backward (which were combinations of sides and directions).

But it might take the rest of the day to type all those doc strings and make sure the tests still pass.

Affie:
> Thanks for the help on the doc strings...

I hope I didn't mess with your local changes, I had a few minutes and wanted to look at your new docs, so just fixed the typos to build the preview docs. I'll look at your new docs, but it might take some time for me to fully understand.

Thanks for taking all the time to clear up the confusion.

kellertuer:
No worries, I could take them easily over to my temporary string storage.

It also helped me to clear up some confusion. I hope the new doc strings and implementations will make that more accessible than the old ideas from GroupManifolds while offering the same amount of features. We'll see.

kellertuer:
The compose docs and implementations is now updated, see

 [this compose and its 7 siblings below](https://juliamanifolds.github.io/LieGroups.jl/previews/PR67/groups/semidirect_product_group/#Manifolds.compose-Union{Tuple{AO},%20Tuple{A},%20Tuple{O2},%20Tuple{O1},%20Tuple{𝔽},%20Tuple{LieGroup{𝔽,%20LeftSemidirectProductGroupOperation{O1,%20O2,%20A,%20AO},%20%3C:ManifoldsBase.ProductManifold},%20Any,%20Any}}%20where%20{𝔽,%20O1,%20O2,%20A%3C:AbstractLeftGroupActionType,%20AO%3C:ActionActsOnLeft})

for all 8 cases. Currently it is also 8 implementation, but I have an idea to recuse it to 4; since 1&5 only differ In all indices 1|2 being switched and we have a common prequel that can also provide those in variables.

The challenge now is that not only `inv` has to be adapted, but also both `diff_left_compose` and `diff_right_compose` have to be generalised to the 8 cases. So I will check really first how to reduce the amount of definitions maybe even back down to not only 4 but 2 (hopefully those that need inv vs those that do not) to reduce amount of possible error in copy paste.

The good thing: After about a year after directions/sides were introduced for GroupManifolds, I now understand them, though I hope I have a slightly better naming scheme here now.

kellertuer:
Here is a state of this PR

* [x] generalise semidirect product to all 8 cases
* [x] mention the new acts_on thingy in the docs and that it replaces `Side` (transition.md)
* [x] improve  docs on left and right actions in transition.md 
* [x] reimplement `compose` accordingly
* [x] fix/reimplement `inv` accordingly
* [x] reimplement (derive/compute before) the differentials of compose (left/right argument)
* write tests that check the functions work fine with aliasing, namely for
  * [x] `apply!`
  * [x] `compose!`
  * [x] `exp!` 
  * [x] `inv!`
  * ~`diff_left_compose!`~
  * ~`diff_right_compose!`~
* [x]  find nice test cases to check the generic (now improved) semidirect Lie group implementation, for example one covering all 8 cases above.

kellertuer:
All implementations should be done the differentials of compose got a bit technical, so I decided to do them in one run, though that meant I just sat all of today just hacking all 16 cases in formula
(left/right differential, left/right Action, left/right semi product, left/right where they act on). But hopefully found a nice way to reduce it to 4 dispatches (left/right diff and left/right act_on).

Please read the docs carefully and maybe try them out, it might take some time to get a good test for these running. I also will put the docs aside a bit, since by now I am _very_ situation blind to spotting typos on that large set of docs. They are, however, comprehensive I hope. They do document _every_ possible formula in detail. I hope I did not make a mistake in the derivations.

kellertuer:
Hm, status from today: somehow calling `submanifold_component` broke, that I just took over from previous implementations - I do not yet know why it broke. But that is what I wasted my time today on.

_edit:_ Something deep inside the matrix variant is so technical that I can not figure out what causes about fnympsteen ambiguities per function call nor where that come from. Will have to take a break and come back to this later. Some dispatch is heavily broken after introducing the ActionActOn parameter.

kellertuer:
Final update for today. As mentioned in #65 I think I misunderstood the group operation action a bit too broadly. But I only noticed that in the generality we now aim for here.

I at least made the existing one more safe to really only act on itself as a group operation as the name suggests. SE(n) now needs a new action type to work with - that does not use compose ;)

src/groups/semidirect_product_group.jl
```diff
@@ -392,89 +969,97 @@ function get_vector_lie!(
...
+    inv(L::LieGroup{𝔽,<:SemidirectProductGroupOperation{⋆,⋄,<:AbstractLeftGroupActionType,AbstractActionActsOn}}, g)
 
-Compute the inverse element of an element ``g = (g_1, g_2)`` given by
+$(_doc_semidirect_sub_groups) Let ``σ`` denote a left group action. The inverse is the same for both [`ActionActsOnLeft`](@ref) and [`ActionActsOnRight`](@ref).
+
+The inverse for the [`LeftSemidirectProductGroupOperation`](@ref) ``$(_math(:∘))`` on ``G ⋉ H`` is given by
 
 ```math
-g^{-1} = (g_1^{-1}, σ_{g_1^{-1}}g_2).
+(g,h){-1} = $(_tex(:bigl))( g^{-1}, σ_{g^{-1}}(h^{-1}) $(_tex(:bigr))).
```
Affie:
```suggestion
(g,h)^{-1} = $(_tex(:bigl))( g^{-1}, σ_{g^{-1}}(h^{-1}) $(_tex(:bigr))).
```

src/groups/semidirect_product_group.jl
```diff
@@ -392,89 +969,97 @@ function get_vector_lie!(
...
+# 7. Right semidirect, right action, act on left
+"""
+    inv(L::LieGroup{𝔽,<:SemidirectProductGroupOperation{⋆,⋄,<:AbstractRightGroupActionType,AbstractActionActsOn}}, g)
+
+$(_doc_semidirect_sub_groups) Let ``τ`` denote a right group action. The inverse is the same for both [`ActionActsOnLeft`](@ref) and [`ActionActsOnRight`](@ref).
+
+The inverse for the [`LeftSemidirectProductGroupOperation`](@ref) ``$(_math(:∘))`` on ``G ⋉ H`` is given by
+
+```math
+(g,h){-1} = $(_tex(:bigl))( g^{-1}, τ_{g}(h^{-1}) $(_tex(:bigr))).
```
Affie:
```suggestion
(g,h)^{-1} = $(_tex(:bigl))( g^{-1}, τ_{g}(h^{-1}) $(_tex(:bigr))).
```

kellertuer:
Thanks! I also think you were right and the new ActsOn-thingy should be part of the semidirect product operation instead, so it would not “disturb” the existing dispatch for the action. Will check that next.

kellertuer:
Looks a bit better today, I am just not clever enough to update out constants like the `LeftSpecialEuclideanGroupOperation` which seems a bit wrong and hence e.g. `zero_vector` does not find its definition to generate a zero matrix in the matrix representation case. That const and the dispatch looks very very very involved to me and I can not figure out why the dispatch does not find the specified method (yet, after an hour of staring at it). We will see how long I have to stare.

kellertuer:
There is now one test failing testing that log is the inverse of exp, not sure why that is now failing. Docs render nice; so I think we are nearly there - besides the aliasing tests :)

kellertuer:
I have not yet written the aliased tests (but prepared the test suite), but I know already from an hour of finding a small indexing bug, that apply and compose on semidirect can handle aliased input very well, at least on h and k (have to check with g and k still).

_edit:_ But the overall tests pass already again on the new semidirect product and all action on the right work fine.

src/groups/semidirect_product_group.jl
```diff
@@ -392,89 +969,97 @@ function get_vector_lie!(
...
+        SDPG::LieGroup{𝔽, <:SemidirectProductGroupOperation{O1, O2, A, AO}, <:ProductManifold}, k, g
+    ) where {𝔽, O1, O2, A <: AbstractRightGroupActionType, AO <: AbstractActionActsOnType}
+    PM, G, H, a, g_ind, h_ind = _semidirect_parts(SDPG)
+    # (a) compute the inverse in H
+    inv!(H, submanifold_component(SDPG, k, Val(h_ind)), submanifold_component(PM, g, Val(h_ind)))
+    # (b) apply the group action w.r.t. g in G (not yet inverted in place) to the inverse from (b)
+    apply!( # Apply the group action with g1^-1 to g2^-1
+        a,
+        submanifold_component(SDPG, k, Val(h_ind)),
+        submanifold_component(SDPG, k, Val(g_ind)),
```
Affie: I think this is a typo and should maybe be from g
```suggestion
        submanifold_component(SDPG, g, Val(g_ind)),
```

src/groups/semidirect_product_group.jl
```diff
@@ -88,296 +96,865 @@ See [HilgertNeeb:2012; Definition 9.2.22](@cite), first definition for more deta
...
-        G,
-        submanifold_component(SDPG, k, Val(2)),
-        submanifold_component(SDPG, g, Val(2)),
-        submanifold_component(SDPG, h, Val(2)),
-    )
+        SDPG::LieGroup{𝔽, <:SemidirectProductGroupOperation{O1, O2, A, AO}, <:ProductManifold}, k, g, h
+    ) where {𝔽, O1, O2, A <: AbstractGroupActionType, AO <: ActionActsOnRight}
+    PM, G, H, a, g_ind, h_ind = _semidirect_parts(SDPG)
+    # invert for right, copy for left
+    _semidirect_maybe_inv!(a, G, submanifold_component(PM, g, Val(g_ind)), submanifold_component(PM, g, Val(g_ind)))
```
Affie: I haven't quite figured this one out yet, but there is a potential issue with `g` being mutated here. The only argument that should be allowed to be mutated is `k`

`g` will be mutated if this dispatch from ln 227 is called:
```
function _semidirect_maybe_inv!(::GroupAction{<:AbstractRightGroupActionType}, G, k, g)
    return inv!(G, k, g)
end
```

src/groups/semidirect_product_group.jl
kellertuer: Thanks for spotting this is correct. Assume we have the left semidirect one, then (a) computed g_2^-1 (and stores it in k_2) – (b) applies g_1*k_2 (or g_1g_2^-1) and stores it in k_2 as well. (c) computes g_1^-1 and stores it in k_1

(c) has to be done ofter (b) to avoid aliased effects.

src/groups/semidirect_product_group.jl
```diff
@@ -392,89 +969,97 @@ function get_vector_lie!(
...
+    SDPG::LieGroup{𝔽, <:SemidirectProductGroupOperation{O1, O2, A, AO}, <:ProductManifold}, ::Any,
+) where {𝔽, O1, O2, A <: AbstractRightGroupActionType, AO <: AbstractActionActsOnType}
+
+function _inv!(
+        SDPG::LieGroup{𝔽, <:SemidirectProductGroupOperation{O1, O2, A, AO}, <:ProductManifold}, k, g
+    ) where {𝔽, O1, O2, A <: AbstractRightGroupActionType, AO <: AbstractActionActsOnType}
+    PM, G, H, a, g_ind, h_ind = _semidirect_parts(SDPG)
+    # (a) compute the inverse in H
+    inv!(H, submanifold_component(SDPG, k, Val(h_ind)), submanifold_component(PM, g, Val(h_ind)))
+    # (b) apply the group action w.r.t. g in G (not yet inverted in place) to the inverse from (b)
```
kellertuer:
```suggestion
    # (b) apply the group action w.r.t. g_g in G (not yet inverted in place) to the inverse g_h from (a)) that is already in k_h; apply it in-place of k_h
```

src/groups/semidirect_product_group.jl
kellertuer: That I have to check carefully, since that was all I did for the last 3 hours basically, hunting single letter things in this code because tests were failing and sometimes some wrong entries ended up in `k` too early. I think I just changed this. So that might mean we have to allocate here. But for today I fear my brain is not able to look at single letters k,g,h any longer – and sad that this is not yet finished/fixed. It was really a lot of work to look through all that today and I had hoped to have a bit of a break from that then.

Well, that's life I guess.

kellertuer: I tried to fix it but I fear in aliased cases this even gets more complicated and we have to allocate the interims result here.

kellertuer: Ah no, changing this makes the tests fail. And they are: We compare to the matrix case where we know the result, so... this is not the right fix.

kellertuer: Found a fix . or in other words did not want to end the day with that bug. Now either component of k is “saved/copied” if the danger of an alias exists - and only the last steps, when g/h are no longer needed write to k (and potentially ovrwrite g,h).
Without aliased memory, no allocations are necessary.

The tests for ActsOnRight work, the ones for left are missing but the logic is adapted already accordingly. Will extend tests together with the aliased tests in the suite.

kellertuer:
Had a few minutes between meetings here and there today and finished all aliased tests I proposed above. So the only thing missing now is test coverage for the cases where the action acts on the left. I think we can even set this now from Draft to a PR and open it for review from others.

Affie:
> Had a few minutes between meetings here and there today and finished all aliased tests I proposed above. So the only thing missing now is test coverage for the cases where the action acts on the left. 

There is one thing that I'm not understanding now; previously `inv` for semidirect products were not covered by the tests numerically, that's why I added explicit tests. So, should all cases not still be covered with the removal of those tests, or am I missing something that changed with dispatch?

> I think we can even set this now from Draft to a PR and open it for review from others.

Sounds good, and also feel free to rename the PR to something more appropriate as it's more a refactor of semidirect products now.

kellertuer:
> There is one thing that I'm not understanding now; previously `inv` for semidirect products were not covered by the tests numerically

Magic! 🪄✨
Ok, not so much, one can just add a function to the list of functions that shall be tested on a LieGroup; no clue why inv was not included there before. Now it is.

Affie:
I experimented a bit with your new aliasing tests, but I get a failure on compose from this part of the test:
```julia
            # alias right input and output
            k5 = copy(G, g)
            compose!(G, k5, g, k5)
            @test isapprox(G, k5, gg; atol = atol)
```

I'll investigate a bit more.

kellertuer:
Interesting ;) All tests pass also the ones on SE(3) with ArrayPartition so I was expecting them to work by now. But maybe I either missed something or on your Lie group one of the components does not yet fully allow for aliased input? TranslationGroup for example is not 100% there because their group operation (eh also called `+`) causes a bit of trouble.

Affie:
My example is SE(2) with my own action and it is the TranslationGroup component that is wrong. I also see it only breaks with cases where the action acts on the left.

The "magic" why `inv` wasn't covered previously and reason why the current SE(n) tests works might be because inv is specifically implemented here?
https://github.com/JuliaManifolds/LieGroups.jl/blob/989f4eb21a4f092c7cc30a0d8e87764a1fe86bef/ext/LieGroupsRecursiveArrayToolsExt/special_euclidean_group_RAT_ext.jl#L148

kellertuer:
Ah there is also one internal change, where one has to be careful.

Previously we only did that for `compose!` where we handled the identity on that level and then passed to a `_compose!` which is the variant where “identity type has been handled” that the user would just implement for their types. This avoids either ambiguity problems and/or to always implement how compose would act on `Identity` inputs for its 3 arguments.

The same split I introduced for `_inv!`, which made it far easier to implement inv and I could remove quite some code (even for all groups) where the identity was handled previously.

src/groups/semidirect_product_group.jl
```diff
@@ -88,296 +96,885 @@ See [HilgertNeeb:2012; Definition 9.2.22](@cite), first definition for more deta
...
+    if Base.mightalias(k, h) || Base.mightalias(k, g) # h/g may not overlap with k
+        kG = copy(G, submanifold_component(SDPG, k, Val(g_ind))) # to
+    else # if it does not alias, we can just use kG & kH
+        kG = submanifold_component(SDPG, k, Val(g_ind))
+    end
+    # invert hG for right, copy for left
+    # this is inplace if both are not aliased and creates a copy kG otherwise to avoid overwriting hG
+    _semidirect_maybe_inv!(a, G, kG, submanifold_component(SDPG, h, Val(g_ind)))
+    # a) group action  (first to avoid side effects in g, set kH to σ_{kG}(gH), with the above this avoids aliasing
+    apply!(a, kH, kG, submanifold_component(SDPG, g, Val(h_ind)))
```
Affie: Just double check this part please as I'm too tired to know for sure.
If `k === h`, `apply!` will modify `kH` and therefore `k[h_inx]`
So, I think ln 306 needs to also check for `Base.mightalias(k, h)`?

src/groups/semidirect_product_group.jl
```diff
@@ -88,296 +96,885 @@ See [HilgertNeeb:2012; Definition 9.2.22](@cite), first definition for more deta
...
-        submanifold_component(SDPG, g, Val(1)),
-        submanifold_component(SDPG, h, Val(1)),
-    )
-    _compose!(
-        H, submanifold_component(SDPG, k, Val(2)), submanifold_component(SDPG, g, Val(2)), x
-    )
+        SDPG::LieGroup{𝔽, <:SemidirectProductGroupOperation{O1, O2, A, AO}, <:ProductManifold}, k, g, h
+    ) where {𝔽, O1, O2, A <: AbstractGroupActionType, AO <: ActionActsOnLeft}
+    PM, G, H, a, g_ind, h_ind = _semidirect_parts(SDPG)
+    if Base.mightalias(k, g) || Base.mightalias(h, g) # k/h may not overlap with g
```
Affie:
```suggestion
    if Base.mightalias(k, g) || Base.mightalias(h, g) || Base.mightalias(k, h) # k/h may not overlap with g or k with h
```

Maybe a fix ... see next comment

Affie: see #68 for tests that failed without this change.

kellertuer: Hm, I sat down yesterday with pen&paper, drew even memory boxes nd had some energy drinks, but you are right I missed this, it seems.
I wanted to be careful to really only allocate when necessary. 

I am not at my computer, can you check, whether `mightalias` also exists with more than 2 arguments? Then we could replace this with one call. What about line 311? Is that fine then?

And - if we change this here we have to also check below 385ish? Maybe also there it is ... ah yes in 399 we have the same situation, so also that needs to be fixed. 

Could you do that?

Affie: > I am not at my computer, can you check, whether mightalias also exists with more than 2 arguments?

Doesn't look like it.

> Could you do that?

I'll try ... this one already broke my brain and I was only able to find it because of a failed test.

kellertuer: Thanks! And no worries, my brain as well yesterday, and I still did not get it right, and just doing the same edits to the other action side below (about line 385) as well then.

Affie: Strangely the other action side didn't break the tests, but I'll have a look.

kellertuer: Sure. In my head those two are symmetric, so should have the same memory structure, but it might also just be un my head ;)

kellertuer: and currently it seems some of the 8 cases got lost in the merge? (or it is a hiccup of codecov)

Affie: It looks like codecov hiccup, I did try and cover `diff_apply, diff_left_compose, diff_right_compose` but got errors and don't know if it is from me or bugs.

src/groups/semidirect_product_group.jl
```diff
@@ -88,296 +96,885 @@ See [HilgertNeeb:2012; Definition 9.2.22](@cite), first definition for more deta
...
+    XG, XH = submanifold_component(sdpg, X, Val(g_ind)), submanifold_component(sdpg, X, Val(h_ind))
+    gG, gH = submanifold_component(SDPG, g, Val(g_ind)), submanifold_component(SDPG, g, Val(h_ind))
+    hG, hH = submanifold_component(SDPG, h, Val(g_ind)), submanifold_component(SDPG, h, Val(h_ind))
+    # For right actions we have to invert hG (g_2) - this allocates when it has to invert
+    hG_mod = _semidirect_maybe_inv(a, G, hG)
+    # one allocation for applying the action
+    # we need one allocation to compute the action (g_2 to h_1)
+    σg2mh1 = apply(a, hG_mod, gH)
+    # Step 1: Compute the argument for the second summand on H - use the memory of YH for the result
+    diff_group_apply!(A, YH, hG_mod, gH, XH)
```
Affie: I think this should be:
```suggestion
    diff_group_apply!(a, YH, hG_mod, gH, XH)
```

kellertuer:
Ah we even checked those at the same time ;) I fear one of the differentials is not yet correct and I would also cover differentials of SE(n) in matrix form, but those should be the two missing cases.

Affie:
> Ah we even checked those at the same time ;) I fear one of the differentials is not yet correct and I would also cover differentials of SE(n) in matrix form, but those should be the two missing cases.

Haha, yes, I agree that the differentials are not correct yet, but I don't have the knowledge to help here. I made a similar addition to the test as you, but it failed so thought I was doing something wrong.

src/groups/semidirect_product_group.jl
kellertuer: No but then it is not a hiccup, we have to cover those.

But I now understand, your code cove report was green/ok because it was not worse than this PR (and it did not compare to master like this one).

kellertuer:
Oh I only have to think about the right actions one, since there is the inverse in as well, that is missing (so a diff investigation in a chain rule basically).

_Update:_ I narrowed it down, it is the cases of the differentials that have sums, (one left one right diff case) and those suffer from an aliasing with the input Lie algebra vector. So luckily all 8 remaining errors stem from that and ist is just a few lines of code affected. Not yet sure how to fix it though. (aliased YH and XH probably)

kellertuer:
I think this is all I can provide. I do not plan to implement any further features.

I hope the code coverage is ok now, otherwise we have to check what to test, still.

kellertuer:
I added a few old tests back for coverage - they use a pullback based on `diff_left_compose`; I have to compare that to the old form but it seems there might be a small sig error somewhere, so to see test coverage I marked that one test as broken.

test/groups/test_special_euclidean_group.jl
```diff
@@ -388,7 +374,9 @@ using StaticArrays
         # If we apply another pullback, this seems to be right, so we have to check where we accidentally pushforward once too much?
         YL = inverse_retract(G, gL, kL, dirm)
         @test is_point(𝔤, YL; error = :error)
-        @test isapprox(𝔤, XL, YL)
+        # TODO: Check the above inverse retract, it seems either the pullback
+        # or the diff_left_compose (LLR case) has a sign error and gets the H component wrong
+        @test_broken isapprox(𝔤, XL, YL; atol = 1.0e-4, error = :error)
```
kellertuer: This needs to be checked. Possible errors
* `pull_back_tangent` it seems a bit strange and could be documented with a proper formula
*  `diff_compose_left!` since it is used in the previous function. 

The symptom is that in the semidirect h-part instead of 1.0 we get a 1.46 - since that starts from 1.23 is might be a sign error on the h component as well.

kellertuer: I have noch checked all internal functions, all their formulae are correct, So what might be is that the case here (LLR) got somehow confused by me with some other case regarding the differential (or the one of the right compose in the above inverse retracts push forward did) so that the “do not match”. But I would have to check probably all 8! cases again, which I do not feel able to.

Maybe someone with a fresh view might see where I confused 2 of the cases either in diff_left or diff_right compose.

kellertuer: I think I fixed the test but I am not yet sure _why_ that is the fix, since now the formula do not agree with the code. Maybe I really mixed up the formulae / cases somewhere. But this way we can at least check for test coverage already.

kellertuer:
I found my mistake at least and want to propose a solution.

For the differentials of compose I sadly used the tangent space representation of tangent vectors instead of the Lie algebra. To not postpone this PR into an indefinite time:

* [x] I deleted the diff_(left|right)_compose code. There was a prior implementation for one of the 8 cases that was as wrong as the new ones. So we “loose” a feature here, but one that was wrong
* [x] I documented the differentials better on the interface page (also clarifies the fixes in #69 )
* [x] I will delete the corresponding tests which probably now fail
* [x] I will revise the code to more carefully use `D` for the classical differential between tangent spaces and `d` for the Lie algebra one – to not confuse them later
* [x] I have to check how far the push forward / pullback code has to be returned to the original format still – update: They indeed work fine on other manifolds, so they seem correct now.
* [x] fixed push forward and pullback which had a wrong global default that is now removed. A default on multiplication groups is provided instead.

Sorry for that long detour into code that in the end was total bogus. Since with #70 it seems not even the original intention of fixing inv here was correct – so the correctness level of this PR just before the last commit was about 0%

Some day when I wrap my had around how this looks like on the Lie algebra I might restart the compose differentials, but I can not yet say how long that takes to understand how that happens on semidirect products.

src/groups/circle_group_complex.jl
```diff
@@ -20,17 +20,19 @@ _doc_diff_left_compose_complex_circ = """
...
 ```
 
-This can be computed in-place of `Y` if `Y` is `mutable`.
+This can be computed in-place of `Y`.
 """
 
 @doc "$(_doc_diff_left_compose_complex_circ)"
 function diff_left_compose(::_ComplexCircleGroup, g::Number, h::Any, X::Number)
-    return g * X
+    return g * X * g^(-1)
```
Affie: Maybe this should be `return X` because the group is Abelian?

kellertuer: ah yes, then it is in deed also just `X`.

src/group_operations/multiplication_operation.jl
```diff
@@ -162,15 +174,20 @@ diff_left_compose(::LieGroup{𝔽, <:AbstractMultiplicationGroupOperation}, g, h
 function diff_left_compose!(
         G::LieGroup{𝔽, <:AbstractMultiplicationGroupOperation}, Y, g, h, X
     ) where {𝔽}
-    return copyto!(LieAlgebra(G), Y, g * X)
+    return copyto!(LieAlgebra(G), Y, h * (X / h))
```
Affie: Is this not `h⁻¹Xh`?

kellertuer: ah yes, but then we can write `(h\X)*h`, which is even a bit nicer.

kellertuer: (the old one `gX` besides a mixup in g and h has by the way the same mistake I made in semidirect - it would be the one between the tangent spaces - well, well...)

Affie:
Just for consistency, the argument order in `diff_right_compose` docs and signatures are sometimes `(g, h)` and sometimes `(h, g)`. 
The arguments are never used so I guess it doesn't really matter, but `(g, h)` makes sense for me with the new docstrings.

kellertuer:
it should always be g,h I can check the occurrences for that again later today.

src/group_operations/addition_operation.jl
```diff
@@ -96,8 +96,13 @@ _doc_diff_right_compose_add = """
     diff_right_compose(G::LieGroup{𝔽,AdditionGroupOperation}, h, g, X)
     diff_right_compose!(G::LieGroup{𝔽,AdditionGroupOperation}, Y, h, g, X)
```
Affie:
```suggestion
    diff_right_compose(G::LieGroup{𝔽,AdditionGroupOperation}, g, h, X)
    diff_right_compose!(G::LieGroup{𝔽,AdditionGroupOperation}, Y, g, h, X)
```

src/group_operations/multiplication_operation.jl
```diff
@@ -162,15 +174,20 @@ diff_left_compose(::LieGroup{𝔽, <:AbstractMultiplicationGroupOperation}, g, h
 function diff_left_compose!(
         G::LieGroup{𝔽, <:AbstractMultiplicationGroupOperation}, Y, g, h, X
     ) where {𝔽}
-    return copyto!(LieAlgebra(G), Y, g * X)
+    return copyto!(LieAlgebra(G), Y, (h \ X) * h)
 end
 
 _doc_diff_right_compose_mult = """
     diff_right_compose(G::LieGroup{𝔽,<:AbstractMultiplicationGroupOperation}, h, g, X)
     diff_right_compose!(G::LieGroup{𝔽,<:AbstractMultiplicationGroupOperation}, Y, h, g, X)
```
Affie:
```suggestion
    diff_right_compose(G::LieGroup{𝔽,<:AbstractMultiplicationGroupOperation}, g, h, X)
    diff_right_compose!(G::LieGroup{𝔽,<:AbstractMultiplicationGroupOperation}, Y, g, h, X)
```

src/interface.jl
```diff
@@ -385,9 +424,29 @@ _doc_diff_right_compose = """
     diff_right_compose(G::AbstractLieGroup, h, g, X)
     diff_right_compose!(G::AbstractLieGroup, Y, h, g, X)
```
```suggestion
    diff_right_compose(G::AbstractLieGroup, g, h, X)
    diff_right_compose!(G::AbstractLieGroup, Y, g, h, X)
```

Affie (commented):
Consistently write diff_right_compose with argument order g,h in docs

kellertuer:
I carefully went through all ~100 occurrences and fixed the order. It did not affect functionality, since it was either in the docs or due to the identity, neither g nor h were used.

Thanks for spotting.

kellertuer (approved):
The docs look fine to me, the tests pass, so maybe this is good to go then

Affie:
This looks good to me as well.

PR #68, opened 2025-08-28
Test 8 Semidirect Product combinations, add compose associativity test, and fix compose alias bug

Affie:
@kellertuer, I made this PR against #67 (note base branch) in case you want to add this way of testing the 8 combinations of Semidirect product groups. 
Feel free to just close this if you have other tests in mind.

### DIFF
--- a/test/LieGroupsTestSuite.jl/LieGroupsTestSuite.jl
+++ b/test/LieGroupsTestSuite.jl/LieGroupsTestSuite.jl
@@ -146,6 +146,12 @@ function test_compose(
     )
     @testset "compose" begin
         k1 = compose(G, g, h)
+
+        #test associativity
+        a1 = compose(G, compose(G, g, k1), h)
+        a2 = compose(G, g, compose(G, k1, h))
+        @test isapprox(G, a1, a2; atol = atol)
+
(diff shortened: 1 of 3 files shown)


kellertuer:
Feel free to close this and merge it to the other branch - just also delete the asserts please. We usually assume all input is valid - if one is unsure, use the ValidationLieGroup. That yields less clutter in the code. But coverage looks very nice!

test/LieGroupsTestSuite.jl/LieGroupsTestSuite.jl
```diff
@@ -146,6 +146,12 @@ function test_compose(
     )
     @testset "compose" begin
         k1 = compose(G, g, h)
+
+        #test associativity
```
kellertuer, line 150: Oh, this is also a neat addition. Exactly the type of tests I like, short and checking an abstract property we expect to hold - but still being aware of approx results.

kellertuer:
Ah its already late you want to merge this into the other PR, sure, after deleting the asserts and running Runic, which maybe you did, its fine to merge this over to the other PR.

Thanks for working on code cov.

PR #69, opened 2025-09-02
Test that `diff_right/left_compose` is on the Lie Algebra and maybe a fix

Affie:
From what I understand tangent vectors should always be represented by their corresponding representants in the Lie algebra.  This PR tests that and fixes `AbstractMultiplicationGroupOperation` based on Manifolds.jl
- `translate_diff!(..., ::LeftForwardAction)`  (`Tuple{LeftAction, LeftSide}`) --- `diff_left_compose`
- `translate_diff!(..., ::RightBackwardAction)` (`Tuple{RightAction, RightSide}`) --- `diff_right_compose`


src/group_operations/multiplication_operation.jl
```diff
@@ -162,7 +162,7 @@ diff_left_compose(::LieGroup{𝔽, <:AbstractMultiplicationGroupOperation}, g, h
 function diff_left_compose!(
         G::LieGroup{𝔽, <:AbstractMultiplicationGroupOperation}, Y, g, h, X
     ) where {𝔽}
-    return copyto!(LieAlgebra(G), Y, g * X)
+    return copyto!(LieAlgebra(G), Y, X)
```
Affie, line 177: https://github.com/JuliaManifolds/Manifolds.jl/blob/a4834aedcc7f7d74b981a3f8dd2e819536040300/src/groups/group.jl#L1038-L1049

src/group_operations/multiplication_operation.jl
```diff
@@ -182,7 +182,7 @@ diff_right_compose(
 function diff_right_compose!(
         G::LieGroup{𝔽, <:AbstractMultiplicationGroupOperation}, Y, g, ::Any, X
     ) where {𝔽}
-    return copyto!(LieAlgebra(G), Y, X * g)
+    return copyto!(LieAlgebra(G), Y, inv(G, g) * X * g)
```
Affie: https://github.com/JuliaManifolds/Manifolds.jl/blob/a4834aedcc7f7d74b981a3f8dd2e819536040300/src/groups/group.jl#L1074-L1085

src/group_operations/multiplication_operation.jl
kellertuer: I am very unsure about this and have to think how the `g` would vanish here and why. And we probably should document that then much more carefully, for best of cases with a reference.

The classical differential is of course `gX`, so maybe we miss an adjoint here? Or to be precise a conjugate?

src/group_operations/multiplication_operation.jl
kellertuer: Same as about, I would think this would generically be a `conjugate` to compute then? But this is a major change so we should also very carefully document it.

src/group_operations/multiplication_operation.jl
Affie: Will the manifolds documentation be sufficient? https://juliamanifolds.github.io/Manifolds.jl/stable/manifolds/group/#Default-Representation-of-Tangent-Vectors

It just seems the other way around to me?

src/group_operations/multiplication_operation.jl
Affie: Yes, Manifolds.jl calls `adjoint_action` by default and the same default implantation can maybe work in LieGroups?

kellertuer: Well, besides that we have renamed it, and it is just called adjoint, probably, yes https://juliamanifolds.github.io/LieGroups.jl/stable/interface/group/#Base.adjoint-Tuple{AbstractLieGroup,%20Any,%20Any}

Maybe this adjoint is also what is missing in the end in the semidirect product. Still, also explaining that in the docs would be best – so that any user can also see _why_ the code is what it is (or the results then).

src/group_operations/multiplication_operation.jl
kellertuer: Check that carefully, as far as I see there, we are interested in `Y` so I think it is not even the other way around - at least only looking at `Y`.

kellertuer: but it might be worth adding this to the interface/generics docs of `diff_left_compose` and `diff_right_compose`, maybe even in this PR since it handles those two.

kellertuer: So I checked this carefully and the proposed fix is not correct here, because that would mean the composition had the identity as derivative and that does not make sense. So something is missing here, maybe also an adjoint?

kellertuer: Ah no, I see now.. I add the documentation on the other branch (since I just had that checked out. Your fixes are correct.

kellertuer: Eh yes it is the other way around than in your code. I now checked the original derivation and added it to the docs. Please also update the docs here accordingly then and not just the code.

Affie: Yes, it looks like it was swopped then. The `LeftForward` `RightBackward` was a bit confusing.

src/group_operations/multiplication_operation.jl
```diff
@@ -153,7 +153,7 @@ _doc_diff_left_compose_mult = """
     diff_left_compose!(G::LieGroup{𝔽,<:AbstractMultiplicationGroupOperation}, Y, g, h, X)
 
 Compute the differential of the left group multiplication ``λ_g(h) = g$(_math(:∘))h``,
-which simplifies for an [`AbstractMultiplicationGroupOperation`](@ref) to ``$(_math(:d))λ_g(h)[X] = $(_math(:Ad))(g)[X] = gXg^{-1}``.
+which simplifies for an [`AbstractMultiplicationGroupOperation`](@ref) to ``$(_math(:d))λ_g(h)[X] = $(_math(:Ad))(h^{-1})[X] = h^{-1}Xh``.
```
Affie: From the interface docs for `diff_left_compose!`:
`dλg(h)[X]=h⁻¹Xh`

kellertuer: Oh, thanks for being careful, I have to check on which of the two updates I made the error then.

Affie:
I might have found a derivation we can use for the left-action semidirect product diff_compose [here](https://web.williams.edu/Mathematics/it3/texts/semidirect.pdf)

kellertuer:
I only see the (quite involved) conjugate? Also most of that is nice mathematically but I am not yet able to match that to code.

src/group_operations/multiplication_operation.jl
kellertuer: Now I am confused, both the interface docs
https://juliamanifolds.github.io/LieGroups.jl/previews/PR69/interface/group/#LieGroups.diff_left_compose!-Tuple{AbstractLieGroup,%20Vararg{Any,%204}}
and the ones here for the group op
https://juliamanifolds.github.io/LieGroups.jl/previews/PR69/interface/operations/#LieGroups.diff_left_compose!-Union{Tuple{𝔽},%20Tuple{LieGroup{𝔽,%20var%22#s24%22,%20M}%20where%20{var%22#s24%22%3C:AbstractMultiplicationGroupOperation,%20M%3C:ManifoldsBase.AbstractManifold{𝔽}},%20Vararg{Any,%204}}}%20where%20𝔽
state the same formula as you write. So – what is then wrong here?

Affie: Note: you are comparing against this PR where i updated the docs already. 
Compare against https://juliamanifolds.github.io/LieGroups.jl/previews/PR67/interface/operations
67 vs 69

kellertuer: Ah, thanks. Now I have to check really carefully again, since I think indeed the g should stay it is a constant, ah yes I think I might have mixed left and right up. Note that for the left compose g is the constant (on the left hand side) and hence should stay, h is the parameter and should become the X.

Will try to find time to fix that here then. Well-spotted. And yeah, I do make too many mistakes on all this topic. Sorry for that.

Affie: I'm currently really confused with all the left right options. I think then the original order was the correct one. 

Does `diff_left_compose` mean with respect to the left argument, ie. pq wrt. p?

Affie: Ah, reading your comment carefully it sounds like it is the differential of left compose, that will make the original order correct.

src/interface.jl
```diff
@@ -391,14 +391,14 @@ For example on matrix manifolds this means, we use ``X ∈ 𝔤`` and hence ``W
 The (classical) differential ``$(_math(:D))λ_g(h): T_g$(_math(:G))) → T_{g$(_math(:∘))h}$(_math(:G)))`` reads
 
 ```math
-  $(_math(:D))λ_g(h)[W] = Wh = gXh = gh(h^{-1}Xh) = gh$(_math(:Ad))(h)[X] ∈ T_{gh}$(_math(:G))).
+  $(_math(:D))λ_g(h)[W] = Wh = gXh = gh(h^{-1}Xh) = gh$(_math(:Ad))(h^{-1})[X] ∈ T_{gh}$(_math(:G))).
```
```suggestion
  $(_math(:D))λ_g(h)[W] = Wh = gXh = gh(h^{-1}Xh) = gh$(_math(:Ad))(h)[X] ∈ T_{gh}$(_math(:G))).
```

src/interface.jl
```diff
@@ -391,14 +391,14 @@ For example on matrix manifolds this means, we use ``X ∈ 𝔤`` and hence ``W
...
 ```
 
 To also bring this back to the Lie algebra, we just have to multiply the result with ``(gh)^{-1}`` from the left
 and obtain
 
 ```math
-  $(_math(:d)) λ_g(h)[X] = h^{-1}Xh = $(_math(:Ad))(h)[X] ∈ 𝔤,
+  $(_math(:d)) λ_g(h)[X] = h^{-1}Xh = $(_math(:Ad))(h^{-1})[X] ∈ 𝔤,
```
```suggestion
  $(_math(:d)) λ_g(h)[X] = h^{-1}Xh = $(_math(:Ad))(h)[X] ∈ 𝔤,
```

src/group_operations/multiplication_operation.jl
Affie: I swopped it back to the original order

kellertuer: The formulae are mixed up not the names, _l_ambda is _L_eft rho is _r_ight.

Affie:
The base branch was fixed. So closing.

PR #70, opened 2025-09-02
Test Semidirect Product against expected `inv` results and fix inv/compose combinations

Affie:
- Adds expected to `test_inv` and tests expected values obtained from matrix inverse against SemidirectProduct groups PR
- Fixed swoped left-right group action on ActionActsOnLeft for `compose` and `inv`.


kellertuer:
I thought it would be enough to test that `compose(G, inv(G,g), g)` is (at least approx) e. But sure, we can do the expected ones as well, only that I would prefer only having these additionally, since they would usually be computed by hand and hence contain “magic numbers” in the sense of “trust me these are correct, I computed that by hand”.

src/groups/semidirect_product_group.jl
```diff
@@ -313,9 +313,15 @@ function _compose!(
...
-    # invert hG for right, copy for left
+    # invert hG for left, copy for right
     # this is inplace if both are not aliased and creates a copy kG otherwise to avoid overwriting hG
-    _semidirect_maybe_inv!(a, G, kG, submanifold_component(SDPG, h, Val(g_ind)))
+    # _semidirect_maybe_inv!(a, G, kG, submanifold_component(SDPG, h, Val(g_ind)))
+    if A <: AbstractLeftGroupActionType
+        inv!(G, kG, submanifold_component(SDPG, h, Val(g_ind)))
+    else
+        copyto!(G, kG, submanifold_component(SDPG, h, Val(g_ind)))
+    end
```
kellertuer: Why this change?
I was hoping the dispatch would make the code clearer and do the same?
So this (and the same below) seem to me as being the equivalent code, but in a worse more crowded code style.

src/groups/semidirect_product_group.jl
```diff
@@ -1022,9 +1033,44 @@ inv(
...
+#         _inv_semidirect_same!(SDPG, k, g)
+#     else
+#         _inv_semidirect_opposite!(SDPG, k, g)
+#     end
+# end
+
+function _inv!(
+        SDPG::LieGroup{𝔽, <:SemidirectProductGroupOperation{O1, O2, A, AO}, <:ProductManifold}, k, g
+    ) where {𝔽, O1, O2, A <: AbstractLeftGroupActionType, AO <: ActionActsOnRight}
+    return _inv_semidirect_forward_action!(SDPG, k, g)
```
kellertuer: I never understood forward, so please let's not use this artificially introduced name here. It is not properly defined (in like I have zero references for this) and it just confuses readers. What did you actually change here?

src/groups/semidirect_product_group.jl
```diff
@@ -1022,9 +1033,44 @@ inv(
...
+end
+
 function _inv!(
         SDPG::LieGroup{𝔽, <:SemidirectProductGroupOperation{O1, O2, A, AO}, <:ProductManifold}, k, g
-    ) where {𝔽, O1, O2, A <: AbstractLeftGroupActionType, AO <: AbstractActionActsOnType}
+    ) where {𝔽, O1, O2, A <: AbstractRightGroupActionType, AO <: ActionActsOnRight}
+    return _inv_semidirect_inverse_action!(SDPG, k, g)
+end
+
+function _inv_semidirect_forward_action!(SDPG, k, g)
```
kellertuer: why do you introduce 3 cases that all call just this one, it has a strange name, is not documented and does the same as before?

That is - to be honest – the same as above: Running the same code logic in less structured, less documented, strange naming.

src/groups/semidirect_product_group.jl
```diff
@@ -1065,9 +1111,10 @@ inv(
     SDPG::LieGroup{𝔽, <:SemidirectProductGroupOperation{O1, O2, A, AO}, <:ProductManifold}, ::Any,
 ) where {𝔽, O1, O2, A <: AbstractRightGroupActionType, AO <: AbstractActionActsOnType}
 
-function _inv!(
-        SDPG::LieGroup{𝔽, <:SemidirectProductGroupOperation{O1, O2, A, AO}, <:ProductManifold}, k, g
-    ) where {𝔽, O1, O2, A <: AbstractRightGroupActionType, AO <: AbstractActionActsOnType}
+function _inv_semidirect_inverse_action!(SDPG, k, g)
```
kellertuer: Now I am a bit more lost. In Manifolds.jl we had forward/backward – here it is then `forward/inverse`? Then I am totally lost in naming, which is when I usually proposed Terry Pratchetts withershins and clockwise as additional namings. Also this is again not documented – why the additional layer of internal functions? The above 4 cases could have stayed the two from before.

Affie: Let me get rid of the bad naming first. It was just a PR for discussion, but it's getting in the way.

src/groups/semidirect_product_group.jl
Affie: Same, I'll get rid of the bad naming first

src/groups/semidirect_product_group.jl
Affie: Same, I'll get rid of the bad naming first

src/groups/semidirect_product_group.jl
Affie: Left/right is swopped between the 2 `_semidirect_maybe_inv` calls, so one function didn't work for this discussion PR.

src/groups/semidirect_product_group.jl
Affie: In the meantime, this is what the idea was about:
```
if AbstractLeftGroupActionType && ActionActsOnRight call _inv_semidirect_forward_action!  #do not inv g
if AbstractRightGroupActionType && ActionActsOnLeft call _inv_semidirect_forward_action!  #do not inv g
if AbstractLeftGroupActionType && ActionActsOnLeft call _inv_semidirect_inverse_action!  #with inv g
if AbstractRightGroupActionType && ActionActsOnRight call _inv_semidirect_inverse_action!  #with inv g
```

src/groups/semidirect_product_group.jl
kellertuer: you can surely dispatch on the different types still, maybe with 4 small cases.

Affie: Sure, if this PR is not a dead end, I will clean this up further.

Affie: The docs also need to be updated then.

Affie:
> I thought it would be enough to test that compose(G, inv(G,g), g) is (at least approx) e. But sure, we can do the expected ones as well, only that I would prefer only having these additionally, since they would usually be computed by hand and hence contain “magic numbers” in the sense of “trust me these are correct, I computed that by hand”.

Ok, I can remove that if you want, but we need something for common mode errors in `compose` and `inv` so I added 2 test cases to compare against the matrix `inv` as sanity checks.

kellertuer:
> > I thought it would be enough to test that compose(G, inv(G,g), g) is (at least approx) e. But sure, we can do the expected ones as well, only that I would prefer only having these additionally, since they would usually be computed by hand and hence contain “magic numbers” in the sense of “trust me these are correct, I computed that by hand”.
> 
> Ok, I can remove that if you want, but we need something for common mode errors in `compose` and `inv` so I added 2 test cases to compare against the matrix `inv` as sanity checks.

Comparing against a carefully matrix inv sanity check sounds very reasonable. we can surely keep those and als the expected idea. Just - if possible – avoid “magic number tests” in the sense of “these are computed by hand, trust me”. Because if such a test fails, someone will quite likely sit down and compute them again. But if you have an alternative route, like the matrix inv, that is of course much better.

src/groups/semidirect_product_group.jl
Affie: I moved this to a common function.

Affie:
Hopefully this is cleaner now for the discussion. What is still outstanding is updating the docstrings. I also don't like the duplication of `_inv`, `_semidirect_maybe_inv!` should also work.

src/groups/semidirect_product_group.jl
kellertuer: I think at least the expected test is a reasonable idea, for the remaining changes, I am not yet sure how different your variant now in the end is. I have not yet checked that.

Affie:
@kellertuer is the logic in this code snippet correct and how do you feel about including it for a generic fallback (if correct)? The idea is that you don't have to implement `inv` of `AbstractGroupActionType`, but doing so saves in calculating `inv(A.group, g)`. I think it can simplify the code in this PR.

```julia
# generic inverse actions fallback implemented with α_g(p)⁻¹ = α_g⁻¹(p)
struct InverseLeftGroupAction{T} <: AbstractRightGroupActionType end
struct InverseRightGroupAction{T} <: AbstractLeftGroupActionType end

Base.inv(act::AbstractRightGroupActionType) = InverseRightGroupAction{typeof(act)}()
Base.inv(act::AbstractLeftGroupActionType) = InverseLeftGroupAction{typeof(act)}()
Base.inv(::InverseRightGroupAction{T}) where {T} = T()
Base.inv(::InverseLeftGroupAction{T}) where {T} = T()

function apply!(A::GroupAction{<:Union{<:InverseRightGroupAction, <:InverseLeftGroupAction}}, k, g, p)
    return apply!(inv(A), k, inv(A.group, g), p)
end
```

```julia
julia> G = SpecialOrthogonalGroup(2);
julia> M = TranslationGroup(2);

julia> α = GroupAction(G, M, LeftMultiplicationGroupAction())
GroupAction(SpecialOrthogonalGroup(2), TranslationGroup(2; field=ℝ), LeftMultiplicationGroupAction())

julia> iα = inv(α)
GroupAction(SpecialOrthogonalGroup(2), TranslationGroup(2; field=ℝ), LieGroups.InverseLeftGroupAction{LeftMultiplicationGroupAction}())

julia> g = rand(G)
2×2 Matrix{Float64}:
 -0.993875   0.110514
 -0.110514  -0.993875

julia> p = rand(M)
2-element Vector{Float64}:
  1.7422423456861844
 -1.837042748529893

julia> apply(α, g, p) ≈ apply(iα, inv(G, g), p)
true
```

kellertuer:
Sure, why not. Took me a while to see that inv is a wrapper, but that follows the same idea as the `Transpose` wrapper in LinearAlgebra then.

src/groups/semidirect_product_group.jl
```diff
@@ -224,11 +224,17 @@ function _semidirect_maybe_inv(::GroupAction{<:AbstractLeftGroupActionType}, G,
...
+function _semidirect_maybe_inv!(::Type{AO}, ::GroupAction{A}, G, k, g) where {AO, A}
+    if (A <: AbstractLeftGroupActionType) != (AO == ActionActsOnLeft)
+        # act on left && right action ||
+        # act on right && left action
+        return copyto!(G, k, g)
+    else
+        # act on left && left action ||
+        # act on right && right action
+        return inv!(G, k, g)
+    end
```
kellertuer, line 231: Previously this was done via dispatch. Now it is an if. Why?

Affie: I found it easier to understand with one function and an if statement than with 4 when using dispatch.

kellertuer: But the if statement is also very cryptic. Is is basically if we have a left group action _and_ it acts on right – as well as we have a right group action and it acts on left?
Why is the first a `<:` and the second a `==`?

Maybe `(A <: AbstractLeftGroupAction) == (AO <: ActionActsOnLeft)` (and switch your two blocks of course) Is easier to parse (for humans)?

It still for 100% changes the logic, since the old one had one case for left actions one for right actions. So if the docs are not updated – they are wrong.

Affie: Look at the comments in each case.
```julia
if  act_on_left && right_action || act_ on_right && left_action 
    return copyto!(G, k, g)
elseif  act_on_left && left_action || act_on_right && right_action
    return inv!(G, k, g)
end
```
This is the core of the change of when to call inverse, it doesn't only depend on `act_on_left` or `act_on_right`, but the combination.

PR #71, opened 2025-10-02
Add entry for transition from `SpecialEuclideanInGeneralLinear`

mateuszbaran:


mateuszbaran:
I'm not sure why the new row doesn't appear in the rendered docs?

kellertuer (approved):

kellertuer:
Oh that is indeed surprising, to me it seems the next two lines are missing as well?

kellertuer:
Fixed it, seems it was a caching issue. The transition.md is within the tutorials so that is cached, and somehow it did then take the one from the cache. It should be fine now.

mateuszbaran:
Thanks!

PR #82, opened 2025-11-24
Fix typo in documentation

wuzh07:


src/Lie_algebra/Lie_algebra_interface.jl
```diff
@@ -371,7 +371,7 @@ _doc_vee = """
     vee!(𝔤::LieAlgebra, c, X)
 
 Compute the vee map ``(⋅){\\vee}: $(_math(:𝔤)) → $(_tex(:Cal, "V"))`` that maps a tangent vector `X`
-from the [`LieAlgebra`](@ref) $(_math(:𝔤)) to its coordinates with respect to the [`DefaultLieAlgebraOrthogonalBasis`](@ref) basis in the Lie algebra
+from the [`LieAlgebra`](@ref) ``𝔤`` to its coordinates with respect to the [`DefaultLieAlgebraOrthogonalBasis`](@ref) basis in the Lie algebra
```
kellertuer:
```suggestion
from the [`LieAlgebra`](@ref) ``$(_math(:𝔤))`` to its coordinates with respect to the [`DefaultLieAlgebraOrthogonalBasis`](@ref) basis in the Lie algebra
```

We use the internal `_math` to keep a persistent notation for the generic notations, so the correct fix here is “just” to add the math environment ``` ``...`` ```, but to keep the notation wrapper as well :)

kellertuer (approved):
I did a minor fix to your fix (see explanation), but thanks for noticing both typos!

PR #89, opened 2026-01-13
Bump Readme

kellertuer:
This PR is just a bit of nice small changes

* add the new paper to the readme
* restructure the badges the same way they are organised in Manifolds.jl
* add a section that quotes both BibLaTeX citations

And this also fixes a small error we had in the docs, since one extref was broken, when the DefaultMetric moved from Manifolds.jl to ManifoldsBase.jl


mateuszbaran (approved):

kellertuer:
Thanks for the help :)

Issue #94, opened 2026-04-17
CircleGroup StackOverflow on inv


dehann:
Hi, this looks similar to:
- #62 

```julia
using LieGroups

M = CircleGroup()
inv(M, [0.0])
```

Results in stack overflow:
```julia
julia> inv(M, [0.0])
Warning: detected a stack overflow; program state may be corrupted, so further execution might be unreliable.
ERROR: StackOverflowError:
Stacktrace:
       [1] GenericMemory
         @ ./boot.jl:588 [inlined]
       [2] Array
         @ ./boot.jl:648 [inlined]
       [3] similar
         @ ./array.jl:404 [inlined]
       [4] allocate
         @ ~/.julia/packages/ManifoldsBase/0PEsI/src/ManifoldsBase.jl:92 [inlined]
       [5] allocate
         @ ~/.julia/packages/ManifoldsBase/0PEsI/src/ManifoldsBase.jl:109 [inlined]
       [6] allocate_result
         @ ~/.julia/packages/ManifoldsBase/0PEsI/src/ManifoldsBase.jl:195 [inlined]
       [7] allocate_result
         @ ~/.julia/packages/LieGroups/MuHqn/src/interface.jl:1266 [inlined]
       [8] _inv(G::LieGroup{ℂ, AbelianMultiplicationGroupOperation, Manifolds.Circle{ℂ}}, g::Vector{Float64})
         @ LieGroups ~/.julia/packages/LieGroups/MuHqn/src/interface.jl:626
       [9] inv(G::LieGroup{ℂ, AbelianMultiplicationGroupOperation, Manifolds.Circle{ℂ}}, g::Vector{Float64})
         @ LieGroups ~/.julia/packages/LieGroups/MuHqn/src/interface.jl:624
      [10] _inv!(G::LieGroup{ℂ, AbelianMultiplicationGroupOperation, Manifolds.Circle{ℂ}}, h::Vector{Float64}, g::Vector{Float64})
         @ LieGroups ~/.julia/packages/LieGroups/MuHqn/src/group_operations/multiplication_operation_abelian.jl:243
 ┌    [11] inv!(G::LieGroup{ℂ, AbelianMultiplicationGroupOperation, Manifolds.Circle{ℂ}}, h::Vector{Float64}, g::Vector{Float64})
 │       @ LieGroups ~/.julia/packages/LieGroups/MuHqn/src/interface.jl:632
[...]

PR #95, opened 2026-04-18
Fix issue #94

mateuszbaran:
This mirrors the way inv is handled in the real circle.


kellertuer:
Hm, code-wise this does something. Semantically I am not so happy.

Here is my problem: Usually we say, that the Lie group (or manifold) exists in an abstract way, maybe remembering its field it works over for technical reasons.
Now your test creates the Circle group for the complex case (the circle in the complex plain), where I would expect points to be (1-element arrays or 0-length arrays of) complex numbers.
But your test uses a real number.

mateuszbaran:
OK, I've changed the test to use complex numbers.

kellertuer (approved):
Now it looks correct.

PR #97, opened 2026-04-24
Fix allocation in the `SpecialEuclideanGroup` constructor

Affie:
Strangely, this caused dynamic dispatch and other allocations in the `TranslationGroup` and `RotationsGroup` constructors that doubled my solve time with RLM, but this fixed cleared that up.
Before and after:
```julia
julia> @btime SpecialEuclideanGroup(2)
  1.577 μs (15 allocations: 720 bytes)
SpecialEuclideanGroup(2)

julia> @btime SpecialEuclideanGroup(2)
  1.148 ns (0 allocations: 0 bytes)
SpecialEuclideanGroup(2)
```

### DIFF
--- a/src/groups/special_euclidean_group.jl
+++ b/src/groups/special_euclidean_group.jl
@@ -178,7 +178,7 @@ end
 function SpecialEuclideanGroup(n::Int; variant::Symbol = :left, kwargs...)
     SOn = SpecialOrthogonalGroup(n; kwargs...)
     Tn = TranslationGroup(n; kwargs...)
-    variant ∉ SA[:left, :right] && error(
+    variant ∉ (:left, :right) && error(
         "SE(n) requires a  variant ∉ [:left, :right] but you provided `variant=:$variant`",
     )
     return variant === :left ? SOn ⋉ Tn : Tn ⋊ SOn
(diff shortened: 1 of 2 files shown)


kellertuer:
Oh, we never were super eager to reduce constructors, since usually you should be able to reuse a manifold/group.

We could also do that line without any array by writing

```
(variant != :left) && (variant != :right) && error(" [...]
```

If you like we can also make this a new version 0.1.11

src/groups/special_euclidean_group.jl
```diff
@@ -178,7 +178,7 @@ end
 function SpecialEuclideanGroup(n::Int; variant::Symbol = :left, kwargs...)
     SOn = SpecialOrthogonalGroup(n; kwargs...)
     Tn = TranslationGroup(n; kwargs...)
-    variant ∉ [:left, :right] && error(
+    variant ∉ SA[:left, :right] && error(
```
mateuszbaran:
```suggestion
    variant ∉ (:left, :right) && error(
```
I think using a tuple would be nicer here and just as fast.

mateuszbaran:
As Ronny wrote, we generally avoid using manifold/group constructors in hot path but it's still good to have this faster.

src/groups/special_euclidean_group.jl
Affie: You are right, I first tried a tuple but must have made a silly mistake because it works.

Affie:
We rely on constant propagation for the Manifolds/Groups (from `Pose{n}` `Point{n}`, etc) and I think the allocations prevented it.

Affie:
> If you like we can also make this a new version 0.1.11

You can wait for more fixes if you want, I'm happy using the slightly slower version or just the main branch for now.

mateuszbaran (approved):

kellertuer (approved):

PR #103, opened 2026-07-13
Add jacobian_exp implementations for SE(2/3) and SGal(3)

Affie:
Hi, this PR adds `jacobian_exp` for SE(2/3), SGal(3), ProductLieGroups, and Abelian Groups.
I did use claude quite a bit especially for the documentation and tests. 

We use it here (this file also contains some fall-backs and matrix group and algebra adjoints that might be worth upstreaming):
https://github.com/JuliaRobotics/ApproxManifoldProducts.jl/blob/b16ea75a69e051374b58064159ad732d9c74f18b/src/services/ManifoldsOverloads.jl#L88

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,13 +9,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-* `jacobian_exp` for the `TranslationGroup` (identity), the `SpecialEuclideanGroup` in
-  both variants and dimensions 2 and 3, and the `SpecialGalileanGroup(3)`
-  (closed form from Kelly, arXiv:2312.07555, section 8). All closed forms fall back to
-  numerically robust expansions near zero rotation angles.
-* `jacobian_exp` for a (direct) `ProductLieGroup`, assembled block-diagonally from the
-  `jacobian_exp` of its factors.
+* `jacobian_exp` for the `TranslationGroup` (identity), the `SpecialEuclideanGroup` in both variants and dimensions 2 and 3, and the `SpecialGalileanGroup(3)` (closed form from Kelly, arXiv:2312.07555, section 8). All closed forms fall back to numerically robust expansions near zero rotation angles.
+* `jacobian_exp` for a (direct) `ProductLieGroup`, assembled block-diagonally from the `jacobian_exp` of its factors.
 * Fix ambiguities with `ZeroVector`and `Identity{AdditionGroupOperation}` in addition operations. `ZeroVector` was introduced in ManifoldsBase v2.5.
+
+### Changed
+
+* `jacobian_exp(G, g, X, b)` and `jacobian_exp!(G, J, g, X, b)` no longer take the base point `g`, since the Jacobian of the group exponential map does not depend on it: use `jacobian_exp(G, X, b)` and `jacobian_exp!(G, J, X, b)` instead. The old signatures are deprecated and still work.
 
 ## [0.1.11] 2026-05-05
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,17 @@ All notable Changes to the Julia package `LieGroups.jl` will be documented in th
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.1.12] unreleased
+
+### Added
+
+* `jacobian_exp` for the `TranslationGroup` (identity), the `SpecialEuclideanGroup` in
+  both variants and dimensions 2 and 3, and the `SpecialGalileanGroup(3)`
+  (closed form from Kelly, arXiv:2312.07555, section 8). All closed forms fall back to
+  numerically robust expansions near zero rotation angles.
+* `jacobian_exp` for a (direct) `ProductLieGroup`, assembled block-diagonally from the
+  `jacobian_exp` of its factors.
+
 ## [0.1.11] 2026-05-05
 
 ### Added
--- a/NEWS.md
+++ b/NEWS.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   numerically robust expansions near zero rotation angles.
 * `jacobian_exp` for a (direct) `ProductLieGroup`, assembled block-diagonally from the
   `jacobian_exp` of its factors.
+* Fix ambiguities with `ZeroVector`and `Identity{AdditionGroupOperation}` in addition operations. `ZeroVector` was introduced in ManifoldsBase v2.5.
 
 ## [0.1.11] 2026-05-05
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,7 +5,7 @@ All notable Changes to the Julia package `LieGroups.jl` will be documented in th
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [0.1.12] unreleased
+## [0.1.12] 2026-08-24
 
 ### Added
 
--- a/Project.toml
+++ b/Project.toml
@@ -22,7 +22,7 @@ LieGroupsTestExt = "Test"
 [compat]
 LinearAlgebra = "1.10"
 Manifolds = "0.11"
-ManifoldsBase = "2"
+ManifoldsBase = "2.5"
 Quaternions = "0.7.6"
 Random = "1.10"
 RecursiveArrayTools = "2, 3, 4.0"
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "LieGroups"
 uuid = "6774de46-80ba-43f8-ba42-e41071ccfc5f"
 authors = ["Seth Axen <seth.axen@gmail.com>", "Mateusz Baran <mateuszbaran89@gmail.com>", "Ronny Bergmann <git@ronnybergmann.net>", "Yueh-Hua Tu", "Olivier Verdier <olivier.verdier@gmail.com>"]
-version = "0.1.11"
+version = "0.1.12"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/ext/LieGroupsRecursiveArrayToolsExt/special_galilean_group_RAT_ext.jl
+++ b/ext/LieGroupsRecursiveArrayToolsExt/special_galilean_group_RAT_ext.jl
@@ -194,13 +194,26 @@ function LieGroups.compose(::SpecialGalileanGroup, g::ArrayPartition, h::ArrayPa
     )
 end
 
-# Lie bracket [X, Y] on 𝔰𝔤𝔞𝔩(3): the matrix commutator of the 5×5 screw representation
-# (see the `hat` docstring), expressed in the ((Ω, ν), (ρ, ι)) blocks
-# [Kelly:2025; eq. (14) basis](@cite):
-#   [X, Y]_Ω = Ω_X Ω_Y - Ω_Y Ω_X   (equivalently ω_X × ω_Y)
-#   [X, Y]_ν = Ω_X ν_Y - Ω_Y ν_X
-#   [X, Y]_ρ = Ω_X ρ_Y - Ω_Y ρ_X + ι_Y ν_X - ι_X ν_Y
-#   [X, Y]_ι = 0
+_doc_lie_bracket_SGal3 = """
+    lie_bracket(𝔰𝔤𝔞𝔩::LieAlgebra{ℝ,<:LeftSpecialGalileanGroupOperation,<:SpecialGalileanGroup}, X, Y)
+    lie_bracket!(𝔰𝔤𝔞𝔩::LieAlgebra{ℝ,<:LeftSpecialGalileanGroupOperation,<:SpecialGalileanGroup}, Z, X, Y)
+
+Compute the Lie bracket ``[X, Y] = XY - YX`` of two tangent vectors `X`, `Y` of the Lie algebra of the
+[`SpecialGalileanGroup`](@ref)`(3)`, i.e. the matrix commutator of their ``5×5`` `hat` representations.
+
+In the ``((\\Omega, \\nu), (\\rho, \\iota))`` block form (see [`hat`](@ref)) this reduces to
+```math
+[X, Y] = \\bigl(
+(\\Omega_X \\Omega_Y - \\Omega_Y \\Omega_X,\\ \\Omega_X \\nu_Y - \\Omega_Y \\nu_X),\\
+(\\Omega_X \\rho_Y - \\Omega_Y \\rho_X + \\iota_Y \\nu_X - \\iota_X \\nu_Y,\\ 0)
+\\bigr).
+```
+The basis is defined in eq. (14) of [Kelly:2025](@cite).
+
+This can be computed in-place of `Z`.
+"""
+
+"$(_doc_lie_bracket_SGal3)"
 function LieGroups.lie_bracket(
         ::typeof(LieAlgebra(SpecialGalileanGroup(3))),
         X::ArrayPartition,
@@ -220,6 +233,7 @@ function LieGroups.lie_bracket(
     )
 end
 
+"$(_doc_lie_bracket_SGal3)"
 function LieGroups.lie_bracket!(
         ::typeof(LieAlgebra(SpecialGalileanGroup(3))),
         Z::ArrayPartition,
--- a/ext/LieGroupsRecursiveArrayToolsExt/special_galilean_group_RAT_ext.jl
+++ b/ext/LieGroupsRecursiveArrayToolsExt/special_galilean_group_RAT_ext.jl
@@ -346,7 +346,5 @@ function LieGroups.jacobian_exp!(
     ρ = X.x[2].x[1]
     ι = X.x[2].x[2][]
     ω = [Ω[3, 2], Ω[1, 3], Ω[2, 1]]
-    # jacobian_exp is the left-trivialized differential of exp (the right Jacobian),
-    # obtained from the left Jacobian of [Kelly:2025, eq. (31)] as J_r(ξ) = J_ℓ(-ξ)
     return LieGroups._jacobian_exp_left_SGal3!(J, -ρ, -ν, -ω, -ι)
 end
--- a/ext/LieGroupsRecursiveArrayToolsExt/special_galilean_group_RAT_ext.jl
+++ b/ext/LieGroupsRecursiveArrayToolsExt/special_galilean_group_RAT_ext.jl
@@ -324,7 +324,6 @@ end
 function LieGroups.jacobian_exp!(
         ::LieGroups.SpecialGalileanGroup{ManifoldsBase.TypeParameter{Tuple{3}}},
         J::AbstractMatrix,
-        g,
         X::ArrayPartition,
         ::DefaultLieAlgebraOrthogonalBasis,
     )
--- a/ext/LieGroupsRecursiveArrayToolsExt/special_galilean_group_RAT_ext.jl
+++ b/ext/LieGroupsRecursiveArrayToolsExt/special_galilean_group_RAT_ext.jl
@@ -1,12 +1,6 @@
-using LieGroups: SpecialGalileanGroup
+using LieGroups: SpecialGalileanGroup, _skew
 using StaticArrays
 using LinearAlgebra
-
-# Internal function to compute the skew-symmetric matrix as an SMatrix used for performance.
-# Can be replaced with hat(SO(3), v) once that works without allocations.
-function _skew(v::AbstractVector{T}) where {T <: Real}
-    return SMatrix{3, 3, T}(0, v[3], -v[2], -v[3], 0, v[1], v[2], -v[1], 0)
-end
 
 # Internal function to compute the matrix Q used in the exponential and logarithm maps for the Special Galilean group. (D matrix in Kelly:2025)
 function _Q(θ⃗)
@@ -200,6 +194,47 @@ function LieGroups.compose(::SpecialGalileanGroup, g::ArrayPartition, h::ArrayPa
     )
 end
 
+# Lie bracket [X, Y] on 𝔰𝔤𝔞𝔩(3): the matrix commutator of the 5×5 screw representation
+# (see the `hat` docstring), expressed in the ((Ω, ν), (ρ, ι)) blocks
+# [Kelly:2025; eq. (14) basis](@cite):
+#   [X, Y]_Ω = Ω_X Ω_Y - Ω_Y Ω_X   (equivalently ω_X × ω_Y)
+#   [X, Y]_ν = Ω_X ν_Y - Ω_Y ν_X
+#   [X, Y]_ρ = Ω_X ρ_Y - Ω_Y ρ_X + ι_Y ν_X - ι_X ν_Y
+#   [X, Y]_ι = 0
+function LieGroups.lie_bracket(
+        ::typeof(LieAlgebra(SpecialGalileanGroup(3))),
+        X::ArrayPartition,
+        Y::ArrayPartition,
+    )
+    ΩX, νX, ρX, ιX = X.x[1].x[1], X.x[1].x[2], X.x[2].x[1], X.x[2].x[2][1]
+    ΩY, νY, ρY, ιY = Y.x[1].x[1], Y.x[1].x[2], Y.x[2].x[1], Y.x[2].x[2][1]
+    return ArrayPartition(
+        ArrayPartition(
+            ΩX * ΩY - ΩY * ΩX,                        # Ω
+            ΩX * νY - ΩY * νX,                         # ν
+        ),
+        ArrayPartition(
+            ΩX * ρY - ΩY * ρX + ιY * νX - ιX * νY,     # ρ
+            zero(X.x[2].x[2]),                        # ι
+        ),
+    )
+end
+
+function LieGroups.lie_bracket!(
+        ::typeof(LieAlgebra(SpecialGalileanGroup(3))),
+        Z::ArrayPartition,
+        X::ArrayPartition,
+        Y::ArrayPartition,
+    )
+    ΩX, νX, ρX, ιX = X.x[1].x[1], X.x[1].x[2], X.x[2].x[1], X.x[2].x[2][1]
+    ΩY, νY, ρY, ιY = Y.x[1].x[1], Y.x[1].x[2], Y.x[2].x[1], Y.x[2].x[2][1]
+    Z.x[1].x[1] .= ΩX * ΩY .- ΩY * ΩX
+    Z.x[1].x[2] .= ΩX * νY .- ΩY * νX
+    Z.x[2].x[1] .= ΩX * ρY .- ΩY * ρX .+ ιY .* νX .- ιX .* νY
+    Z.x[2].x[2] .= 0
+    return Z
+end
+
 # Dev NOTE: hat and vee use a different bases order than that of the underlining semidirect + direct product groups,
 # therefore, get_vector_lie and get_coordinates_lie are implemented explicitly. see hat/vee docstrings for details.
 function LieGroups.get_vector_lie(
@@ -284,4 +319,21 @@ function LieGroups.get_coordinates_lie!(
     c[9] = X.x[1].x[1][2, 1] # θ⃗ₓ[2,1]
     c[10] = X.x[2].x[2][]   # Δt
     return c
+end
+
+function LieGroups.jacobian_exp!(
+        ::LieGroups.SpecialGalileanGroup{ManifoldsBase.TypeParameter{Tuple{3}}},
+        J::AbstractMatrix,
+        g,
+        X::ArrayPartition,
+        ::DefaultLieAlgebraOrthogonalBasis,
+    )
+    Ω = X.x[1].x[1]
+    ν = X.x[1].x[2]
+    ρ = X.x[2].x[1]
+    ι = X.x[2].x[2][]
+    ω = [Ω[3, 2], Ω[1, 3], Ω[2, 1]]
+    # jacobian_exp is the left-trivialized differential of exp (the right Jacobian),
+    # obtained from the left Jacobian of [Kelly:2025, eq. (31)] as J_r(ξ) = J_ℓ(-ξ)
+    return LieGroups._jacobian_exp_left_SGal3!(J, -ρ, -ν, -ω, -ι)
 end
--- a/src/group_operations/addition_operation.jl
+++ b/src/group_operations/addition_operation.jl
@@ -21,6 +21,12 @@ Base.:-(e::Identity{AdditionGroupOperation}, ::Identity{AdditionGroupOperation})
 Base.:-(::Identity{AdditionGroupOperation}, g) = -g
 Base.:-(g, ::Identity{AdditionGroupOperation}) = g
 
+# Identity{AdditionGroupOperation} is a valid point on a Lie algebra and the same as the zero vector.
+Base.:+(e::Identity{AdditionGroupOperation}, ::ManifoldsBase.ZeroVector) = e
+Base.:+(::ManifoldsBase.ZeroVector, e::Identity{AdditionGroupOperation}) = e
+Base.:-(e::Identity{AdditionGroupOperation}, ::ManifoldsBase.ZeroVector) = e
+Base.:-(::ManifoldsBase.ZeroVector, e::Identity{AdditionGroupOperation}) = e
+
 _doc_compose_add = """
     compose(G::LieGroup{𝔽,AdditionGroupOperation}, g, h)
     compose!(G::LieGroup{𝔽,AdditionGroupOperation}, k, g, h)
--- a/src/group_operations/addition_operation.jl
+++ b/src/group_operations/addition_operation.jl
@@ -149,8 +149,8 @@ end
 end
 
 _doc_jacobian_exp_add = """
-    jacobian_exp(G::LieGroup{𝔽,AdditionGroupOperation}, g, X, ::DefaultLieAlgebraOrthogonalBasis)
-    jacobian_exp!(G::LieGroup{𝔽,AdditionGroupOperation}, J, g, X, ::DefaultLieAlgebraOrthogonalBasis)
+    jacobian_exp(G::LieGroup{𝔽,AdditionGroupOperation}, X, ::DefaultLieAlgebraOrthogonalBasis)
+    jacobian_exp!(G::LieGroup{𝔽,AdditionGroupOperation}, J, X, ::DefaultLieAlgebraOrthogonalBasis)
 
 Compute the Jacobian of the Lie group exponential in a basis of the Lie algebra on a
 [`LieGroup`](@ref) with an [`AdditionGroupOperation`](@ref).
@@ -160,11 +160,11 @@ identity, so ``J = I_n`` for every ``X``.
 """
 
 @doc "$(_doc_jacobian_exp_add)"
-jacobian_exp(::LieGroup{𝔽, AdditionGroupOperation}, g, X, basis = DefaultLieAlgebraOrthogonalBasis()) where {𝔽}
+jacobian_exp(::LieGroup{𝔽, AdditionGroupOperation}, X, basis = DefaultLieAlgebraOrthogonalBasis()) where {𝔽}
 
 @doc "$(_doc_jacobian_exp_add)"
 function jacobian_exp!(
-        ::LieGroup{𝔽, AdditionGroupOperation}, J::AbstractMatrix, g, X, ::DefaultLieAlgebraOrthogonalBasis
+        ::LieGroup{𝔽, AdditionGroupOperation}, J::AbstractMatrix, X, ::DefaultLieAlgebraOrthogonalBasis
     ) where {𝔽}
     copyto!(J, LinearAlgebra.I)
     return J
--- a/src/group_operations/addition_operation.jl
+++ b/src/group_operations/addition_operation.jl
@@ -142,6 +142,28 @@ end
     return convert(T, c)
 end
 
+_doc_jacobian_exp_add = """
+    jacobian_exp(G::LieGroup{𝔽,AdditionGroupOperation}, g, X, ::DefaultLieAlgebraOrthogonalBasis)
+    jacobian_exp!(G::LieGroup{𝔽,AdditionGroupOperation}, J, g, X, ::DefaultLieAlgebraOrthogonalBasis)
+
+Compute the Jacobian of the Lie group exponential in a basis of the Lie algebra on a
+[`LieGroup`](@ref) with an [`AdditionGroupOperation`](@ref).
+
+Since such a group is Abelian and flat, the differential of the exponential map is the
+identity, so ``J = I_n`` for every ``X``.
+"""
+
+@doc "$(_doc_jacobian_exp_add)"
+jacobian_exp(::LieGroup{𝔽, AdditionGroupOperation}, g, X, basis = DefaultLieAlgebraOrthogonalBasis()) where {𝔽}
+
+@doc "$(_doc_jacobian_exp_add)"
+function jacobian_exp!(
+        ::LieGroup{𝔽, AdditionGroupOperation}, J::AbstractMatrix, g, X, ::DefaultLieAlgebraOrthogonalBasis
+    ) where {𝔽}
+    copyto!(J, LinearAlgebra.I)
+    return J
+end
+
 _doc_identity_element_add = """
     identity_element(G::LieGroup{𝔽,AdditionGroupOperation})
     identity_element!(G::LieGroup{𝔽,AdditionGroupOperation}, e)
--- a/src/groups/special_euclidean_group.jl
+++ b/src/groups/special_euclidean_group.jl
@@ -823,8 +823,8 @@ _se_translation_first(::LeftSpecialEuclideanGroup) = false
 _se_translation_first(::RightSpecialEuclideanGroup) = true
 
 _doc_jacobian_exp_SE2 = """
-    jacobian_exp(G::SpecialEuclideanGroup{TypeParameter{Tuple{2}}}, g, X, ::DefaultLieAlgebraOrthogonalBasis)
-    jacobian_exp!(G::SpecialEuclideanGroup{TypeParameter{Tuple{2}}}, J, g, X, ::DefaultLieAlgebraOrthogonalBasis)
+    jacobian_exp(G::SpecialEuclideanGroup{TypeParameter{Tuple{2}}}, X, ::DefaultLieAlgebraOrthogonalBasis)
+    jacobian_exp!(G::SpecialEuclideanGroup{TypeParameter{Tuple{2}}}, J, X, ::DefaultLieAlgebraOrthogonalBasis)
 
 Compute the Jacobian of the Lie group exponential in a basis of the Lie algebra on the
 [`SpecialEuclideanGroup`](@ref)`(2)`.
@@ -855,23 +855,22 @@ order ``(θ, x, y)``.
 """
 
 @doc "$(_doc_jacobian_exp_SE2)"
-jacobian_exp(::SpecialEuclideanGroup{ManifoldsBase.TypeParameter{Tuple{2}}}, g, X, basis = DefaultLieAlgebraOrthogonalBasis())
+jacobian_exp(::SpecialEuclideanGroup{ManifoldsBase.TypeParameter{Tuple{2}}}, X, basis = DefaultLieAlgebraOrthogonalBasis())
 
 @doc "$(_doc_jacobian_exp_SE2)"
 function jacobian_exp!(
         G::SpecialEuclideanGroup{<:ManifoldsBase.TypeParameter{Tuple{2}}},
-        J::AbstractMatrix, g, X::AbstractMatrix, ::DefaultLieAlgebraOrthogonalBasis,
+        J::AbstractMatrix, X::AbstractMatrix, ::DefaultLieAlgebraOrthogonalBasis,
     )
     return _jacobian_exp_SE2!(G, J, X)
 end
 function jacobian_exp!(
         G::SpecialEuclideanGroup{<:ManifoldsBase.TypeParameter{Tuple{2}}},
         J::AbstractMatrix,
-        g,
         X::SpecialEuclideanMatrixTangentVector,
         B::DefaultLieAlgebraOrthogonalBasis,
     )
-    return jacobian_exp!(G, J, g, ManifoldsBase.internal_value(X), B)
+    return jacobian_exp!(G, J, ManifoldsBase.internal_value(X), B)
 end
 
 function _jacobian_exp_SE2!(
@@ -917,8 +916,8 @@ function _jacobian_exp_SE2!(
 end
 
 _doc_jacobian_exp_SE3 = """
-    jacobian_exp(G::SpecialEuclideanGroup{TypeParameter{Tuple{3}}}, g, X, ::DefaultLieAlgebraOrthogonalBasis)
-    jacobian_exp!(G::SpecialEuclideanGroup{TypeParameter{Tuple{3}}}, J, g, X, ::DefaultLieAlgebraOrthogonalBasis)
+    jacobian_exp(G::SpecialEuclideanGroup{TypeParameter{Tuple{3}}}, X, ::DefaultLieAlgebraOrthogonalBasis)
+    jacobian_exp!(G::SpecialEuclideanGroup{TypeParameter{Tuple{3}}}, J, X, ::DefaultLieAlgebraOrthogonalBasis)
 
 Compute the Jacobian of the Lie group exponential in a basis of the Lie algebra on the
 [`SpecialEuclideanGroup`](@ref)`(3)`.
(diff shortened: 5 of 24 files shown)


Affie:
Locally I only got 6 ambiguities:
```
Number of LieGroups.jl ambiguities: 6
Test Summary: | Pass  Total     Time
LieGroups.jl  | 5463   5463  3m12.4s
     Testing LieGroups tests passed 
```

kellertuer:
Thanks for the PR!
We had 6 ambiguities before, those are fine, see our bound at https://github.com/JuliaManifolds/LieGroups.jl/blob/1a25650bc7b24e2799b7815e92e687e4e28c2ca6/test/test_aqua.jl#L20

Thanks for the transparency on AI, then I will need more time to check the PR more closely – at lest the reference does exist 😄
In order for the `@cite` to work, though, you have add the corresponding entry also to the https://github.com/JuliaManifolds/LieGroups.jl/blob/main/docs/src/references.bib

_edit:_ Oh I See, on CI it raised to 20; that we should carefully check, why that is the case; It was still 6 on last run of CI here, so I fear it is something in this PR then – and most probably interaction with the `Identity` https://juliamanifolds.github.io/LieGroups.jl/stable/interface/operations/#LieGroups.Identity

Affie:
> Thanks for the transparency on AI, then I will need more time to check the PR more closely – at lest the reference does exist 😄

I really suck with writing documentation, but did explicitly tell AI what references to use for the docs and know them well, (they were already in use as well).

kellertuer:
> (they were already in use as well).

Oh sorry! I was a bit quick there, since I only checked that it is not added in this PR, I did (sadly) not check whether it was already in there.

Yeah I am not a fan of AI, but since I get told at least 3 times a day by now I should do everything with AI and spend all my personal money (no support by university there) on some mega-token plan – I gave up arguing against AI and am just fine when people are honest by now.
Usually I have to compensate other peoples use of AI and fixing their mistakes then, that is what I am a bit tired of – but also nothing I can change. It is what life is by now probably, we are all slaves of AI.
And here it just means it might take a while until I get around to check the formulae against the source and such. At least I see the docs do include formulae, which is what we aim for in our packages.

Affie:
Ahh, I know why I only got 6 ambiguities, my local environment was stil on ManifoldsBase v2.4.0:
```
(LieGroups) pkg> st
Project LieGroups v0.1.12
Status `~/.julia/dev/LieGroups/Project.toml`
  [1cead3c2] Manifolds v0.11.28
⌃ [3362f125] ManifoldsBase v2.4.0
  [94ee1d12] Quaternions v0.7.7
  [90137ffa] StaticArrays v1.9.18
  [37e2e46d] LinearAlgebra v1.12.0
  [9a3f8284] Random v1.11.0
Info Packages marked with ⌃ have new versions available and may be upgradable.
```
If I update to ManifoldsBase v2.5.0 I also get the same failure:
```
Ambiguities: Test Failed at /home/johan/.julia/dev/LieGroups/test/test_aqua.jl:28
  Expression: length(ms) <= LG_LIMIT
   Evaluated: 10 <= 6
```

Affie:
The maths part should be fine, that was mostly me, but I did make some mistakes that was picked up. You will see the most complicated is for SGal and i referenced the equations from https://arxiv.org/pdf/2312.07555v11 directly.
I also checked it using monte carlo against the truncated series of the jacobian (test/groups/jacobian_exp_series_reference.jl) and spot checks are included in the tests. 

Edit: Objective look would be very welcome as I tested it against my own understanding, so not very objective.

kellertuer:
Oh! I remembered that we print the ambiguities, see https://github.com/JuliaManifolds/LieGroups.jl/actions/runs/29268851175/job/86881148111#step:6:316

And my guess was not wrong – it's always the identity – just that this one is on us, it's the new + and - ambiguities with the new ZeroVector type, so its not your fault and you can keep the ambiguity on 10 for now. I hope to remember an issue for that tomorrow then.

Affie:
I think it is complaining about methods like `+(::Identity{AdditionGroupOperation}, Y::ManifoldsBase.ZeroVector)  which is not valid operations. Should we ignore it or maybe throw an ArgumentError?

edit: ok I'll keep on 10 for now.

kellertuer:
Actually it is valid:
Identity{Addition} is a valid point on a Lie algebra (not only on the Lie group with group op +) and indeed that identity _is_ the same as the zero vector (because that _is_ the identity in a vector space).

We can really just define those 4 functions to return the identity argument of the two and we are fine.

Indeed both types have exactly the same reason: we want to avoid allocating zeros.

Affie:
You are correct, I didn't think that far. I'll add those 4 methods tomorrow.

Affie:
Like this:
```
Base.:+(g::Identity{AdditionGroupOperation}, ::ManifoldsBase.ZeroVector) = g
Base.:+(::ManifoldsBase.ZeroVector, g::Identity{AdditionGroupOperation}) = g
Base.:-(g::Identity{AdditionGroupOperation}, ::ManifoldsBase.ZeroVector) = g
Base.:-(::ManifoldsBase.ZeroVector, g::Identity{AdditionGroupOperation}) = g
```

kellertuer:
Yes. Exactly that. Perfect :)
And sure, it is two languages meeting basically, the zero vector was even inspired by the identity, so I think its just fair to resolve that here.

Thanks for adding them even here (and fixing our problem there) – maybe even add a short note to the changelog just for completeness :)

ext/LieGroupsRecursiveArrayToolsExt/special_galilean_group_RAT_ext.jl
```diff
@@ -200,6 +194,47 @@ function LieGroups.compose(::SpecialGalileanGroup, g::ArrayPartition, h::ArrayPa
...
 
+# Lie bracket [X, Y] on 𝔰𝔤𝔞𝔩(3): the matrix commutator of the 5×5 screw representation
+# (see the `hat` docstring), expressed in the ((Ω, ν), (ρ, ι)) blocks
+# [Kelly:2025; eq. (14) basis](@cite):
+#   [X, Y]_Ω = Ω_X Ω_Y - Ω_Y Ω_X   (equivalently ω_X × ω_Y)
+#   [X, Y]_ν = Ω_X ν_Y - Ω_Y ν_X
+#   [X, Y]_ρ = Ω_X ρ_Y - Ω_Y ρ_X + ι_Y ν_X - ι_X ν_Y
+#   [X, Y]_ι = 0
+function LieGroups.lie_bracket(
+        ::typeof(LieAlgebra(SpecialGalileanGroup(3))),
```
kellertuer, line 218:
```suggestion
        ::LieGroups.SpecialGalileanGroup{ManifoldsBase.TypeParameter{Tuple{3}}}),
```
why the `typeof`? I think this is the right type (see also below)?

ext/LieGroupsRecursiveArrayToolsExt/special_galilean_group_RAT_ext.jl
```diff
@@ -200,6 +194,47 @@ function LieGroups.compose(::SpecialGalileanGroup, g::ArrayPartition, h::ArrayPa
...
+        ),
+        ArrayPartition(
+            ΩX * ρY - ΩY * ρX + ιY * νX - ιX * νY,     # ρ
+            zero(X.x[2].x[2]),                        # ι
+        ),
+    )
+end
+
+function LieGroups.lie_bracket!(
+        ::typeof(LieAlgebra(SpecialGalileanGroup(3))),
```
kellertuer, line 238:
```suggestion
        ::LieGroups.SpecialGalileanGroup{ManifoldsBase.TypeParameter{Tuple{3}}},
```

same as in line 205

ext/LieGroupsRecursiveArrayToolsExt/special_galilean_group_RAT_ext.jl
```diff
@@ -285,3 +320,20 @@ function LieGroups.get_coordinates_lie!(
...
+        g,
+        X::ArrayPartition,
+        ::DefaultLieAlgebraOrthogonalBasis,
+    )
+    Ω = X.x[1].x[1]
+    ν = X.x[1].x[2]
+    ρ = X.x[2].x[1]
+    ι = X.x[2].x[2][]
+    ω = [Ω[3, 2], Ω[1, 3], Ω[2, 1]]
+    # jacobian_exp is the left-trivialized differential of exp (the right Jacobian),
```
kellertuer, line 349: I can still not wrap my head around all the left/right things here. left-trivialized is right Jacobian?

For me there exist only one Jacobian: The one from the Lie algebra into the Lie algebra.

Formally one could to that in tangent spaces, but that would complicate things unnecessarily – for example the identity we get below is then suddenly a parallel transport, so Lie groups are nice to have here.

src/group_operations/addition_operation.jl
```diff
@@ -142,6 +142,28 @@ end
...
 
+_doc_jacobian_exp_add = """
+    jacobian_exp(G::LieGroup{𝔽,AdditionGroupOperation}, g, X, ::DefaultLieAlgebraOrthogonalBasis)
+    jacobian_exp!(G::LieGroup{𝔽,AdditionGroupOperation}, J, g, X, ::DefaultLieAlgebraOrthogonalBasis)
+
+Compute the Jacobian of the Lie group exponential in a basis of the Lie algebra on a
+[`LieGroup`](@ref) with an [`AdditionGroupOperation`](@ref).
+
+Since such a group is Abelian and flat, the differential of the exponential map is the
+identity, so ``J = I_n`` for every ``X``.
```
kellertuer, line 159: This would even hold for _any_ basis. But we can keep it limited to the default one for sure.

src/groups/special_euclidean_group.jl
```diff
@@ -817,6 +817,203 @@ function _log_SE3!(G::SpecialEuclideanGroup{ManifoldsBase.TypeParameter{Tuple{3}
...
+_se_translation_first(::RightSpecialEuclideanGroup) = true
+
+_doc_jacobian_exp_SE2 = raw"""
+    jacobian_exp(G::SpecialEuclideanGroup{TypeParameter{Tuple{2}}}, g, X, ::DefaultLieAlgebraOrthogonalBasis)
+    jacobian_exp!(G::SpecialEuclideanGroup{TypeParameter{Tuple{2}}}, J, g, X, ::DefaultLieAlgebraOrthogonalBasis)
+
+Compute the Jacobian of the Lie group exponential in a basis of the Lie algebra on the
+[`SpecialEuclideanGroup`](@ref)`(2)`.
+
+For ``X ∈ 𝔰𝔢(2)`` with translation component ``v = (x, y)`` and rotation angle ``θ``,
```
kellertuer: If you like to use the non raw string somewhere, we have some nice `_tex(...)` functions, especially to keep notation consistent for things like special Lie algebras and groups.

src/groups/special_euclidean_group.jl
```diff
@@ -817,6 +817,203 @@ function _log_SE3!(G::SpecialEuclideanGroup{ManifoldsBase.TypeParameter{Tuple{3}
...
+@doc "$(_doc_jacobian_exp_SE2)"
+jacobian_exp(::SpecialEuclideanGroup{ManifoldsBase.TypeParameter{Tuple{2}}}, g, X, basis = DefaultLieAlgebraOrthogonalBasis())
+
+@doc "$(_doc_jacobian_exp_SE2)"
+function jacobian_exp!(
+        G::SpecialEuclideanGroup{<:ManifoldsBase.TypeParameter{Tuple{2}}},
+        J::AbstractMatrix,
+        g,
+        X::AbstractMatrix,
+        ::DefaultLieAlgebraOrthogonalBasis,
```
kellertuer:
```suggestion
        J::AbstractMatrix, g, X::AbstractMatrix, ::DefaultLieAlgebraOrthogonalBasis,
```

Since we are on Runic, we can put multiple arguments in one line, so you could write this a bit nicer like this
Not all fit on one line, so maybe having the Lie group on its own and these on one line is a bit nicer than the whole vertical thingy with each on its own.

src/groups/special_euclidean_group.jl
```diff
@@ -817,6 +817,203 @@ function _log_SE3!(G::SpecialEuclideanGroup{ManifoldsBase.TypeParameter{Tuple{3}
...
+
+````math
+J = \begin{pmatrix}
+\frac{\sin θ}{θ} & \frac{1-\cos θ}{θ} & W_2 x - W_1 y \\
+-\frac{1-\cos θ}{θ} & \frac{\sin θ}{θ} & W_1 x + W_2 y \\
+0 & 0 & 1
+\end{pmatrix},
+\qquad
+W_1 = \frac{1-\cos θ}{θ^2},\quad W_2 = \frac{θ-\sin θ}{θ^2},
+````
```
kellertuer: How sure are you about these formulae? Do I have to verify them? Do I just have to check whether they render nicely?

kellertuer (commented):
Here is a first few comments. I did not find the time yet to check/render the docs locally,
but the code in general looks fine to me

mateuszbaran:
I'm a bit confused here. Is this the Jacobian of the group exponential or special cases of `jacobian_exp_argument` or `jacobian_exp_basepoint`? You do specify the `g` but don't specify with respect to what the Jacobian is calculated.

kellertuer:
Oh I was confused by that as well, but wanted to first check more closely whether that is on me ;)

To provide detail here: the base point one is the differential with respect to g (X fixed) and the arg one is the one with respect to X (g fixed).
Since the Lie group exponential has a fixed g=e (the identity), the Lie group one only has the argument one – but then no g. Or to even make it more precise, X is then also always from the Lie algebra.

Affie:
> .... Since the Lie group exponential has a fixed g=e (the identity), the Lie group one only has the argument one – but then no g. Or to even make it more precise, X is then also always from the Lie algebra.

I saw this and followed the existing pattern, eg:
https://github.com/JuliaManifolds/LieGroups.jl/blob/1a25650bc7b24e2799b7815e92e687e4e28c2ca6/src/groups/special_orthogonal_group.jl#L122-L133

I wasn't too sure about the `g` argument though.
If `g` is not at the identity will this be equivalent to `jacobian_exp_argument`?

src/groups/special_euclidean_group.jl
Affie: I checked with the preview docs: https://juliamanifolds.github.io/LieGroups.jl/previews/PR103/groups/special_euclidean_group/#LieGroups.jacobian_exp
They are from Chirikjian p 36:
image

Affie: I'm happy with this one. I just checked it again.

kellertuer: Ah, I did not see, you are a bonus member here even that GitHub trusts to let the docs be build ;) It indeed looks good.

ext/LieGroupsRecursiveArrayToolsExt/special_galilean_group_RAT_ext.jl
Affie: I think this one is wrong, it should be for the Lie Algebra:
```
julia> typeof(LieAlgebra(SpecialGalileanGroup(3)))
ManifoldsBase.Fiber{ℝ, TangentSpaceType, SpecialGalileanGroup{ManifoldsBase.TypeParameter{Tuple{3}}, LeftSemidirectProductGroupOperation{LeftSemidirectProductGroupOperation{MatrixMultiplicationGroupOperation, AdditionGroupOperation, LeftMultiplicationGroupAction, ActionActsOnRight}, ProductGroupOperation{Tuple{AdditionGroupOperation, AdditionGroupOperation}}, LieGroups.RotationBoostAction, ActionActsOnRight}, ProductManifold{ℝ, Tuple{ProductManifold{ℝ, Tuple{Manifolds.Rotations{ManifoldsBase.TypeParameter{Tuple{3}}}, Manifolds.Euclidean{ℝ, ManifoldsBase.TypeParameter{Tuple{3}}}}}, ProductManifold{ℝ, Tuple{Manifolds.Euclidean{ℝ, ManifoldsBase.TypeParameter{Tuple{3}}}, Manifolds.Euclidean{ℝ, ManifoldsBase.TypeParameter{Tuple{1}}}}}}}}, Identity{LeftSemidirectProductGroupOperation{LeftSemidirectProductGroupOperation{MatrixMultiplicationGroupOperation, AdditionGroupOperation, LeftMultiplicationGroupAction, ActionActsOnRight}, ProductGroupOperation{Tuple{AdditionGroupOperation, AdditionGroupOperation}}, LieGroups.RotationBoostAction, ActionActsOnRight}}}
```

ext/LieGroupsRecursiveArrayToolsExt/special_galilean_group_RAT_ext.jl
Affie: Same here.

src/groups/special_euclidean_group.jl
Affie: I'll have to learn how to use that a bit better. Looking at the other doc-strings, i think this should be `𝔤`

kellertuer: That would be the generic Lie algebra, for SE(e) the one you use is right. As long as you use a raw string, your code is also fine. I see where you have the g from, there we were not consistent and it should also be se.

The reason why we usually use this definition
https://github.com/JuliaManifolds/LieGroups.jl/blob/1a25650bc7b24e2799b7815e92e687e4e28c2ca6/src/documentation_glossary.jl#L179

to interpolate into a string with `$(_math(:se))` is that this is a semantic description. If we ever decide to change the way se looks like in the docs, we just have to change that line 179 above.

But I do see that means one has to change several things in your doc string so if you prefer to leave it as it is now (please stay with se then not g) – that is also fine. I can refactor that later somewhen.

ext/LieGroupsRecursiveArrayToolsExt/special_galilean_group_RAT_ext.jl
kellertuer: Ah yes, reading. Sometimes not so easy :D sure it is of course the Lie algebra and hence a bit more involved to write. I think I would prefer the longer version. If you feel that is too clumsy, keep the typeof.

ext/LieGroupsRecursiveArrayToolsExt/special_galilean_group_RAT_ext.jl
Affie: Yes, this is a big pain point and you previously also made a note [here](https://juliamanifolds.github.io/LieGroups.jl/previews/PR103/notation/#About-left-and-right-Jacobians). I left it in here to make the link to the literature. I prefer the simplicity LieGroups.jl brings. The way I understand it:

_In LieGroups.jl, we use a left-trivialized convention for tangent vectors. This choice means the differential of the exponential map corresponds to a right-sided perturbation, which estimation literature (e.g., Chirikjian, Solà) explicitly defines as the Right Jacobian (J_r)._

kellertuer: > The way I understand it:

My problem here is: I neither really understand left/right-trivialized, not left/right perturbed.

For me, what we do here is: We have the (Lie) exponential map exp that maps from the Lie algebra into the Lie group.
its differential hence maps from the Lie algebra (in short: the tangent space of the tangent space is again the tangent space, so the domain is unchanged) to the Lie algebra.

In this formulation (maybe a bit of a mathy one, ok) – no left/right shenanigans is needed ;)

ext/LieGroupsRecursiveArrayToolsExt/special_galilean_group_RAT_ext.jl
Affie: I don't have a preference, I just do not know how to write it properly, is it just the first part?

kellertuer: What I meant is, you can indeed just keep your typeof. I feel it is a bit vague (since it is implicit), but the explicit one is very very long and even harder to read. So for readability I think the vagueness of the typeof is ok.

mateuszbaran:
I see, I forgot that already exists. I've checked the discussion in https://github.com/JuliaManifolds/LieGroups.jl/pull/65 and I have no clue why the `_argument` suffix was dropped but the point argument was not. Either both should be dropped or both kept. Maybe this is a good time to make this straight. @kellertuer , is there a reason behind this convention or was that an oversight? If it's the second case, do you prefer adding `_argument` or dropping `p`/`g`?

kellertuer:
Hm, I mean this https://github.com/JuliaManifolds/LieGroups.jl/pull/65#issuecomment-3338556463 definitely says that we only have exp_arg here, so there is no need to do the base one – and to then pass on the the manifold (arg) jac exp?.

I think what we could do, is the same as the default described herre
https://juliamanifolds.github.io/LieGroups.jl/stable/interface/group/#Base.exp-Tuple{AbstractLieGroup,%20Any,%20Any}
compared to here
https://juliamanifolds.github.io/LieGroups.jl/stable/interface/group/#Base.exp-Tuple{AbstractLieGroup,%20Any}

and do kind of both – but one for each.
For the first – with g – that defaults to the second anyways, we introduce the arg one – pushing forward the second case – allowing for a nice consistency with manifolds/tangent-spacey X
For the second case we can introduce the drop-g/p case. 

I am just not sure which should be the default to implement, but I think the second one is probably nicer to implement and the first push-forward-defaults.

I hope this is – given the early evening time – not written in too confusing words.

mateuszbaran:
Yes, that comment explains that it's essentially `jacobian_exp_argument`, and it even comments that `g` may not be necessary, so I think it was just an oversight that `g` wasn't removed in that PR.

> and do kind of both – but one for each.
> For the first – with g – that defaults to the second anyways, we introduce the arg one – pushing forward the second case – allowing for a nice consistency with manifolds/tangent-spacey X
> For the second case we can introduce the drop-g/p case.

That makes sense to me.

> I am just not sure which should be the default to implement, but I think the second one is probably nicer to implement and the first push-forward-defaults.

The "no `g`" case as the default to implement sounds good. So basically all this PR needs to do is drop `g` and in the next PR we can fix the other groups I think.

Affie:
> The "no g" case as the default to implement sounds good. So basically all this PR needs to do is drop g and in the next PR we can fix the other groups I think.

The interface and tests are also affected and it looks like only one group was implemented (SO2/3). So it looks easier to drop and deprecate the `g` parameter in this branch. I'm busy removing it for you to look at.

src/interface.jl
```diff
@@ -888,28 +888,40 @@ function jacobian_conjugate!(
...
-    return jacobian_exp!(G, J, g, X, B)
+    J = ManifoldsBase.allocate_result(G, jacobian_exp, X, B)
+    return jacobian_exp!(G, J, X, B)
 end
 
+#TODO deprecated in v0.1.12, remove in v0.2.0
+@deprecate jacobian_exp(
+    G::AbstractLieGroup, g, X, B::AbstractBasis = DefaultLieAlgebraOrthogonalBasis()
+) jacobian_exp(G, X, B)
+
```
Affie, line 934: @mateuszbaran and @kellertuer, this seems like the easiest way to deprecate the `g` argument without breaking changes. Since `jacobian_exp` and the tests included an unused `g` argument, I think it is easiest to change everywhere for `jacobian_exp!` to still work.

src/interface.jl
```diff
@@ -888,28 +888,63 @@ function jacobian_conjugate!(
...
+where ``$(_tex(:rm, "ad"))_X = [X, ⋅]`` denotes the adjoint of the [`LieAlgebra`](@ref), see [`lie_bracket`](@ref).
+The Jacobian ``J`` is the matrix of this map with respect to the basis ``b``: its ``j``th column contains the
+coordinates of ``$(_math(:d))$(_tex(:exp))_{$(_math(:G))}(X)[X_j]``, where ``X_j`` is the ``j``th basis vector of ``b``.
+Since it only depends on ``X``, this Jacobian is independent of a base point, which is why no point is passed.
+
+!!! note
+    In the robotics and state-estimation literature this left-trivialized differential is often called the
+    _right Jacobian_ ``J_r``, for example in [SolaDerayAtchuthan:2021](@cite) and [Chirikjian:2012](@cite).
+    It is related to the _left Jacobian_ ``J_ℓ`` by ``J_r(X) = J_ℓ(-X)``, which is the right-trivialized
+    differential ``$(_tex(:sum))_{k ≥ 0} $(_tex(:frac, "($(_tex(:rm, "ad"))_X)^k", "(k+1)!"))``.
```
Affie, line 919: @kellertuer You helped me figure this out when we strugled with the left and right confusion in `diff_*`. This follows that explanation you wrote in the `diff_*` docstrings, so hopefully it clears it up for users and future me.

Affie: link to the rendered section:
https://juliamanifolds.github.io/LieGroups.jl/previews/PR103/interface/group/#LieGroups.jacobian_exp

ext/LieGroupsRecursiveArrayToolsExt/special_galilean_group_RAT_ext.jl
Affie: I hope this helps: https://github.com/JuliaManifolds/LieGroups.jl/pull/103/changes#r3585626528

src/interface.jl
kellertuer: I would prefer to not use 

>  that is representable as a matrix in a basis of g

A linear map is _always_ representable as a matrix, there is nothing to “turn” 

nor

> left-trivialize

at least not in doc strings, if so only somewhere in a footnote of the docs page.

In order for Future-Johannes to clear that up, present-Ronny will need some time to read the rest very carefully. All your left/right stuff _always_ confuses me. In short: To get form some TgG back to the Lie group, you just have to “pull back” - or in practice multiply by inv(G,g)

kellertuer: on another note, I would not repeat the whole docs / formula for dexp - that has its own implementation and you can refer to the corresponding doc string.

Affie: Yes, the left/right stuff is very confusing, but definitely not mine, that is why I tried to write the comments to trace if the specific literature calls the same thing left or right. I think it is correct for LieGroups.jl to be fixed in one way (left-trivialized) and that is it, but time will tell. 

> or in practice multiply by inv(G,g)

Exactly, but we have 2 options, LieGroups.jl multiplies this "pull back" on the left side. We could have chosen the right side throughout LieGroups.jl as well.

kellertuer: > We could have chosen the right side throughout LieGroups.jl as well.

No. LieGroups.jl is left side land ;)

I personally do not like the term left-trivialised. If at all that is terminology-complification. Lets not use that term here.

src/interface.jl
mateuszbaran: Yes, this looks like the right way.

src/interface.jl
Affie: I’m open to anything. Just tried to follow the sources, but I’m not a mathematician. 
I can delete those lines. What do you call this “LieGroups.jl is left side land”.

src/interface.jl
Affie: Are you sure you don’t mean left 😂

src/interface.jl
kellertuer: zou have a lot of left/right choices actually, I think 3 or 4. The approach in Manifolds.jl had those all in I think. Here we simplified things to mainly always choose the left one, since Hilgert & Neeb did as well.

src/interface.jl
```diff
@@ -888,28 +888,63 @@ function jacobian_conjugate!(
...
+``$(_tex(:exp))_{$(_math(:G))}: $(_math(:𝔤)) → $(_math(:G))`` at ``X ∈ $(_math(:𝔤))``,
+represented in an [`AbstractBasis`](@extref `ManifoldsBase.AbstractBasis`) ``b`` of the [`LieAlgebra`](@ref) ``$(_math(:𝔤))``.
+
+The (classical) differential ``$(_math(:D))$(_tex(:exp))_{$(_math(:G))}(X): $(_math(:𝔤)) → T_{$(_tex(:exp))_{$(_math(:G))}(X)}$(_math(:G))``
+maps a tangent vector of the Lie algebra to a tangent vector at the point ``$(_tex(:exp))_{$(_math(:G))}(X)``.
+To turn this into a map ``$(_math(:𝔤)) → $(_math(:𝔤))``, that is representable as a matrix in a basis of ``$(_math(:𝔤))``,
+we _left-trivialize_ it: analogous to [`diff_right_compose`](@ref) we “pull back” the resulting tangent vector
+by multiplying with ``$(_tex(:exp))_{$(_math(:G))}(X)^{-1}`` from the left.
+The resulting left-trivialized differential ``$(_math(:d))$(_tex(:exp))_{$(_math(:G))}(X): $(_math(:𝔤)) → $(_math(:𝔤))``
+has the series representation
```
```suggestion
To express this as a map on the Lie algebra ``$(_math(:𝔤)) → $(_math(:𝔤))``, which (in the finite-dimensional case) can be represented as a matrix in a basis of ``$(_math(:𝔤))``,
we push forward the output tangent vector from the tangent space back to the Lie algebra via the differential of the left translation by ``$(_tex(:exp))_{$(_math(:G))}(X)^{-1}``. 
For a matrix Lie group, this action corresponds to multiplying the tangent vector from the left.
The resulting map ``$(_math(:d))$(_tex(:exp))_{$(_math(:G))}(X): $(_math(:𝔤)) → $(_math(:𝔤))``
has the convergent series representation
```

src/interface.jl
mateuszbaran: You're correct, I keep confusing left and right around Lie groups :laughing:

src/interface.jl
kellertuer: yes, that is fine with me. I would use pull back as the actual term personally and put “left translate” in quotes instead, but maybe that is just personal favour

src/interface.jl
mateuszbaran: > A linear map is _always_ representable as a matrix, there is nothing to “turn”

I think it's worth mentioning that this is only trivial in the finite-dimensional case :wink: .

> > We could have chosen the right side throughout LieGroups.jl as well.
> 
> No. LieGroups.jl is left side land ;)

We tried not sticking to the left-side convention in Manifolds.jl before and it's a nightmare to keep it consistent. Left-side it slightly more performance-friendly in some cases so I think it makes sense to go with it, even if the right side would be more consistent with what Manifolds.jl is doing outside of Lie groups.

src/interface.jl
Affie: I think i see the issue with left translate; translate is on the group and we are doing a push-forward from TₚG to 𝔤. Then pull back might be a bit misleading as well as pull-back is normally for differential forms, correct?
So, I've updated this part again, what do you think now?
I'm not a mathematician and suck with documentation, so feel free to change anything in this PR.

kellertuer: ok.

ext/LieGroupsRecursiveArrayToolsExt/special_galilean_group_RAT_ext.jl
```diff
@@ -200,6 +194,47 @@ function LieGroups.compose(::SpecialGalileanGroup, g::ArrayPartition, h::ArrayPa
...
 end
 
+# Lie bracket [X, Y] on 𝔰𝔤𝔞𝔩(3): the matrix commutator of the 5×5 screw representation
+# (see the `hat` docstring), expressed in the ((Ω, ν), (ρ, ι)) blocks
+# [Kelly:2025; eq. (14) basis](@cite):
+#   [X, Y]_Ω = Ω_X Ω_Y - Ω_Y Ω_X   (equivalently ω_X × ω_Y)
+#   [X, Y]_ν = Ω_X ν_Y - Ω_Y ν_X
+#   [X, Y]_ρ = Ω_X ρ_Y - Ω_Y ρ_X + ι_Y ν_X - ι_X ν_Y
+#   [X, Y]_ι = 0
+function LieGroups.lie_bracket(
```
kellertuer, line 217: Is this somewhere documented, what the formula is and that it is available on SGal3?

Also, we do not need to implement this twice, `lie_bracket` should allocate its `Z` and call `lie_bracket!`.
If this does not work “out of the box” (without default implementation outside this ext) – the function here should allocate the specific Z necessary and call the function in line 223

Affie: No I didn't document it separately, it is from the normal matrix formula $$[X, Y] = XY - YX$$. The 2 implementations was to support StaticArrays. I still not know what pattern to follow there other than explicit implementations.

Thinking of it now, we should implement `lie_bracket` for semi-direct groups and then this `lie_bracket` can call the generic semi-direct one.

Affie: It is tested indirectly through https://github.com/JuliaManifolds/LieGroups.jl/pull/103/changes#diff-a2df98ae7a1a8ffb49dc810ab63cf5324c21ffc62aed8cbb0e18bcc78191ec29R21 to match the jacobian.

kellertuer: Sure, we always test both :) The question is mainly whether it has to be implemented twice.

As for the documentation. Ok; I do not have a super good overview, but I did notice that here https://juliamanifolds.github.io/LieGroups.jl/stable/groups/special_galilean_group/ one can not see that the function is available. That is where the question about docs comes from; but if the formula is super default, I am not so sure what best to do.

Affie: I cannot find this docstring in the rendered docs:

https://github.com/JuliaManifolds/LieGroups.jl/blob/1a25650bc7b24e2799b7815e92e687e4e28c2ca6/ext/LieGroupsRecursiveArrayToolsExt/special_euclidean_group_RAT_ext.jl#L252-L268

I do not think special_euclidean_group_RAT_ext.jl docs are rendered.

kellertuer: some things in extensions are weird. One of them is docs I fear. 
We do not yet have doc strings attached in extensions I think. I would not know how to get that to work in Documenter.

kellertuer: Ah, nice fix :)

Project.toml
dehann, line 3:
```suggestion
authors = ["Seth Axen <seth.axen@gmail.com>", "Mateusz Baran <mateuszbaran89@gmail.com>", "Ronny Bergmann <git@ronnybergmann.net>", "Yueh-Hua Tu", "Olivier Verdier <olivier.verdier@gmail.com>", "NavAbility(TM) Contributors <info@wherewhen.ai>"]
```

dehann: Hi, could we be included in this list as minor authors please?  We use the general "Contributors" handle to keep it simple, e.g. the code here has a pretty deep origin story (and is already covered by proper contributor agreements to ensure safe public release).  LieGroups.jl is definitely the right place for it, and there probably are a few more general functions to upstream after this.

kellertuer: No, sorry – I disagree here
That list is not the Contributors but the maintainers. Yes, neither Seth nor Olivier are current maintainers, but both were. For example when this was still in manifolds.jl a large part of this code was developed and maintained by Seth.

Anyone in this list (unless reported absent) I would expect to react to issues and PR and fix them and they are maintainers of the Repo.

But we have the Metadata file that is for example used to generate the Zenodo Authors list

https://github.com/JuliaManifolds/LieGroups.jl/blob/main/.zenodo.json

Also there I would prefer real names actually, but there, everyone doing a PR is invited to add them if they feel it fits.

kellertuer: Also I do not understand the part “is already covered by proper contributor agreements to ensure safe public release” – what does that even mean?

kellertuer: ...oh, adendum: Besides the formal addition to the zenodo metadata, feel free to extend the `about.md` page https://juliamanifolds.github.io/LieGroups.jl/stable/about/, where also other contributions like from Paula and Leonard, as well as Olivier (and Johannes already as well) are mentioned.

Affie:
I've genuinely learned a lot working through this and previous PRs with you and think these have been a team effort, with a lot of help along the way from you and Mateusz.

---

Happy to keep going on the remaining review comments whenever you have time.

Project.toml
Affie: Following up since this got resolved without a reply: to be clear, this PR isn't upstreamed content — it's new work written directly for LieGroups.jl. I'll leave the "contributor agreements" comment for Dehann to clarify if he wants to, that's not something I can speak to.

Affie:
> For this PR, the only thing I noticed now in the files I have not yet marked as read: In the extension we now have doc strings. This also would lead to the point they can only “be read” once the extension is loaded.
Still the string itself does not mention the points being something “special” (only available in the extension). Might that confuse users?

Good point on the doc strings. Right now the whole SGal group implementation depends on the RecursiveArrayTools extension (via ArrayPartition) — there's no pure matrix-form implementation yet. So it's less "these specific functions need the extension" and more "SGal itself only works once the extension is loaded."

The SGal group constructor does have a "The ArrayPartition (default) implementation requires RecursiveArrayTools.jl to be loaded. The matrix representation is not implemented yet." line. 

I looked at ProductLieGroup as that also only works with the RAT extension, that throws a clear error when you try to use it: "This error could be resolved by importing RecursiveArrayTools.jl.". The ProductLieGroup doc strings doesn't mention RAT. 

The "This error could be resolved by importing RecursiveArrayTools.jl." way might be the clearest, but I could not find that yet.

Affie:
> Besides that, we should only remember to set the date in the news before merging the PR.

I saw you normally set the date and bump the version just before release in a separate PR.
PowerLieGroup is also missing `jacobian_exp`, it should not be hard to do and I can add it here or in a new PR if we want to get this one merged.

kellertuer:
There is two ways to provide a good error (besides adding the note to the doc string).

A very clean method is adding error hints like

https://github.com/JuliaManifolds/ManifoldsBase.jl/blob/42478fd08aa0e6f5520091b324af057d4e846290/src/ManifoldsBase.jl#L1285-L1301

and I was surprised we have nearly none for RAT. There we used another trick, probably from before the error hints existed

https://github.com/JuliaManifolds/ManifoldsBase.jl/blob/42478fd08aa0e6f5520091b324af057d4e846290/src/ManifoldsBase.jl#L197-L203

we check whether one allocates on a product manifold hit that default case there. Before you reach this array case, you come from `allocate_on`. loading RAT this is overwritten at

https://github.com/JuliaManifolds/ManifoldsBase.jl/blob/42478fd08aa0e6f5520091b324af057d4e846290/ext/ManifoldsBaseRecursiveArrayToolsExt/ProductManifoldRecursiveArrayToolsExt.jl#L33-L35

This is where you see that message from :) or do not see it after loading

I personally like the hint better since it keeps the original error message.

kellertuer:
> PowerLieGroup is also missing jacobian_exp, 

since we twiddle with that function here, we can also do that here, no need to do a separate PR I think.

src/interface.jl
```diff
@@ -888,28 +888,63 @@ function jacobian_conjugate!(
...
+where ``$(_tex(:rm, "ad"))_X = [X, ⋅]`` denotes the adjoint of the [`LieAlgebra`](@ref), see [`lie_bracket`](@ref).
+The Jacobian ``J`` is the matrix of this map with respect to the basis ``b``: its ``j``th column contains the
+coordinates of ``$(_math(:d))$(_tex(:exp))_{$(_math(:G))}(X)[X_j]``, where ``X_j`` is the ``j``th basis vector of ``b``.
+Since it only depends on ``X``, this Jacobian is independent of a base point, which is why no point is passed.
+
+!!! note
+    In the robotics and state-estimation literature this left-trivialized differential is often called the
+    _right Jacobian_ ``J_r``, for example in [SolaDerayAtchuthan:2021](@cite) and [Chirikjian:2012](@cite).
+    It is related to the _left Jacobian_ ``J_ℓ`` by ``J_r(X) = J_ℓ(-X)``, which is the right-trivialized
+    differential ``$(_tex(:sum))_{k ≥ 0} $(_tex(:frac, "($(_tex(:rm, "ad"))_X)^k", "(k+1)!"))``.
```
mateuszbaran: I'm sorry but I'm confused again: I thought we settled that the Jacobians we have in LieGroups.jl are left Jacobians? Because we keep representing all tangent vectors on Lie groups in the Lie algebra.

Also, where did you get the formula relating left and right Jacobians from? Sola:2021 has a different formula (Eq. (57)) which looks ~right~ correct to me (despite numerous issues in other parts).

src/interface.jl
```diff
@@ -888,28 +888,63 @@ function jacobian_conjugate!(
...
+maps a tangent vector of the Lie algebra to a tangent vector at the point ``$(_tex(:exp))_{$(_math(:G))}(X)``.
+To express this as a map on the Lie algebra ``$(_math(:𝔤)) → $(_math(:𝔤))``, which (in the finite-dimensional case) can be represented as a matrix in a basis of ``$(_math(:𝔤))``,
+we push forward the output tangent vector from the tangent space back to the Lie algebra via the differential of the left translation by ``$(_tex(:exp))_{$(_math(:G))}(X)^{-1}``. 
+For a matrix Lie group, this action corresponds to multiplying the tangent vector from the left.
+The resulting map ``$(_math(:d))$(_tex(:exp))_{$(_math(:G))}(X): $(_math(:𝔤)) → $(_math(:𝔤))``
+has the convergent series representation
+
+```math
+$(_math(:d))$(_tex(:exp))_{$(_math(:G))}(X) = $(_tex(:sum))_{k ≥ 0} $(_tex(:frac, "(-$(_tex(:rm, "ad"))_X)^k", "(k+1)!")),
+```
```
mateuszbaran, line 908: I'd suggest adding a reference for the formula, from a quick search [1], Theorem 5.4 should be good enough.

[1] B. C. Hall, Lie Groups, Lie Algebras, and Representations: An Elementary Introduction, vol. 222. in Graduate Texts in Mathematics, vol. 222. Cham: Springer International Publishing, 2015. doi: [10.1007/978-3-319-13467-3](https://doi.org/10.1007/978-3-319-13467-3).

test/groups/jacobian_exp_series_reference.jl
```diff
@@ -0,0 +1,40 @@
+# Independent series ground truth for `jacobian_exp`, built purely from LieGroups
+# primitives (`lie_bracket`, `hat`, `vee`) so the tests validate the closed forms
+# against math rather than hard-coded magic numbers. Shared by the SE and SGal tests.
+
+using LieGroups, ManifoldsBase
+using LinearAlgebra
+
+if !isdefined(@__MODULE__, :_jacobian_exp_series)
```
mateuszbaran: Why exactly is this check necessary?

mateuszbaran:
Everything else looks fine to me.

src/interface.jl
kellertuer: with “it looks right” you probably mean “looks correct” – just to (not again) open several definitions/interpretations of “right” in parallel 😄

mateuszbaran: Yes, you're correct :laughing:

kellertuer:
Besides Mateusz remarks it also looks good to me. Do not worry about the nightly failing, that seems to be not related to your changes.

test/groups/jacobian_exp_series_reference.jl
Affie: It appears it is not. I wanted to avoid a warning, but I think it was only development related. Will remove.

Affie:
```suggestion
```

src/interface.jl
Affie: >Also, where did you get the formula relating left and right Jacobians from? 

Equation (76) in Sola.

Left/right

The literature was quite confusing on this matter, but I came to the conclusion that Sola and Chirikjian calls this the right Jacobian (Jᵣ) and some sources the same Jacobian left,  Jₗ (sometimes left-trivialized Jacobian). 

Closest reference that explicitly refers to this is from Chirikjian p27:
_15 The “l” and “r” convention used here for Jacobians and for vector ﬁelds is opposite that
used in the mathematics literature. The reason for the choice made here is to emphasize the
location of the “the most informative part” of the expression. In Jacobians, this is the location
of the partial derivatives. In vector ﬁelds, this is where the components deﬁning the ﬁeld appear._

kellertuer: Re Left/Right:

This package follows in notation a naming Hilgert/Neeb. Both are mathematicians. They do not use the term “trivialized” w.r.t. Jacobians; the word Jacobian appears exactly _once_ in that book ;)

Because actually what we even called Jacobian here for completeness is the differential – since exp is on the Lie algebra without the need to distinguish base point / argument – there is only the argument `X`.

mateuszbaran: > Equation (76) in Sola.

OK, I see -- it's a special case of Jacobian of `exp.`

> The literature was quite confusing on this matter, but I came to the conclusion that Sola and Chirikjian calls this the right Jacobian (Jᵣ) and some sources the same Jacobian left, Jₗ (sometimes left-trivialized Jacobian).

This one I don't see though. All relevant literature seems to imply that "left Jacobians" are those that operate on the Lie algebra, and LieGroups.jl also stores tangent vectors in the Lie algebra. What's the reasoning that "right Jacobians" are what we need here?

> This package follows in notation a naming Hilgert/Neeb. Both are mathematicians. They do not use the term “trivialized” w.r.t. Jacobians; the word Jacobian appears exactly _once_ in that book ;)

I'm not a fan of the term "trivialized" either, especially when used without explanation. It's clear, for me, right now, what is meant here but I'd prefer explicitly mentioning composition and conjugation by an element, which are already established terms for what happens here.

Affie: One of my checks were LieGroups.jl [`diff_left_compose` ](https://juliamanifolds.github.io/LieGroups.jl/stable/interface/group/#LieGroups.diff_left_compose-Tuple{AbstractLieGroup,%20Any,%20Any,%20Any}). To get back to the tangent space left multiplication is used resulting in the differential `Ad(h⁻¹)[X]` matching the "right Jacobians" of Sola eq 65.

If I understand Chirikjian 10.4.1 correctly it also matches Jᵣ, with `g⁻¹`multiplied on the left to push-forward the vector from the tangent space to the lie algebra. I have Jₗ as the version that multiplies `g⁻¹` on the right to get back from the tangent space to the Lie algebra (I saw some called this right-trivialized).

> I'm not a fan of the term "trivialized" either...

 I wanted this note to keep track of how different literature matches to LieGroups.jl, I'm open to changing it to reflect the correct terminology, but think it will be helpful to pin down how left/right from the sources align with LieGroups.jl for future use.

src/interface.jl
Affie: I added the reference you suggested.

src/interface.jl
mateuszbaran: > One of my checks were LieGroups.jl [diff_left_compose ](https://juliamanifolds.github.io/LieGroups.jl/stable/interface/group/#LieGroups.diff_left_compose-Tuple%7BAbstractLieGroup,%20Any,%20Any,%20Any%7D). To get back to the tangent space left multiplication is used resulting in the differential Ad(h⁻¹)[X] matching the "right Jacobians" of Sola eq 65.

I see, that's indeed a "right Jacobian" we already have. Well, then it looks OK to me to also provide the "right Jacobian" for `exp`.

> I wanted this note to keep track of how different literature matches to LieGroups.jl, I'm open to changing it to reflect the correct terminology, but think it will be helpful to pin down how left/right from the sources align with LieGroups.jl for future use.

I think it's OK to keep "left/right trivialization" as an alternative terminology but I'd still prefer if the primary explanation avoided this term.

kellertuer:
_moderation note_:
This PR included until a few seconds ago a detour into attribution of contributions. This was now deleted by me.
Reason: A corresponding reference like https://github.com/JuliaRobotics/Caesar.jl/blob/develop/Project.toml does not contain authors, the documentation not a single page about contributions from others and no Zenodo metadata in the repository either. The only place with an `author` is the citation file, also not listing other contributors.

All further comments on this topic will be deleted without further comment.

kellertuer (approved):
I think the only open point is the one note to literature about left/right Jacobians and trivialisations – which opposed to the name – is not a trivial note to write right (=correct).

I do not like those terms, but I do see that there might be a need to add such a note for others that come from the related literature. So for me that looks fine and I hereby already approve the PR; I would prefer also an approval from @mateuszbaran as well though :)

Thanks again @Affie for doing this PR!

mateuszbaran:
> I think the only open point is the one note to literature about left/right Jacobians and trivialisations – which opposed to the name – is not a trivial note to write right (=correct).
> 
> I do not like those terms, but I do see that there might be a need to add such a note for others that come from the related literature. So for me that looks fine and I hereby already approve the PR; I would prefer also an approval from @mateuszbaran as well though :)

After the discussion I don't really have a strong opinion whether we should have left or right Jacobians here. Both make sense, one can be easily obtained from the other and as long as they are properly documented it should be fine. The last outstanding thing is the use of the term "left/right trivializtion" which I'd prefer to make a bit less prominent but it's not a blocking issue for me. The rest looks fine.

mateuszbaran (approved):

kellertuer:
The trivialisation is no longer mentioned much, just once in the docs for people that recognise it, but for example no longer in comments (see last commit).

NEWS.md
```diff
@@ -5,6 +5,18 @@ All notable Changes to the Julia package `LieGroups.jl` will be documented in th
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.1.12] unreleased
```
kellertuer:
```suggestion
## [0.1.12] 2026-08-24
```

Project.toml
kellertuer:
```suggestion
version = "0.1.12"
```

PR #105, opened 2026-09-10
Update CI, Update Docs

kellertuer:
This is the companion PR to e.g. https://github.com/JuliaManifolds/ManifoldsBase.jl/pull/290

* [x] DocumenterCodeBlocks
* [x] DocumenterLandingPage
* [x] a labelto deactivate the changelog check
* there is no label to deactivate the docs, because here they have to be activated anyways
* [x] adapt dependabot to use the new changelog label.

For the landing page I am not yet done with the 6 tiles, they are now autogenerated and I had to remove a total nonsense one. Will need a bit of time and thinking to adopt these.
But besides that, this was at least only 30 minutes.


kellertuer:
I also added the typos/Spellcheck CI – and hopefully found 6 nice tiles on the landing page. This is ready and can be merged when approved.

kellertuer:
I am not sure this is worth its own version. Somewhen – but maybe also only after a few manifolds checked with code analysis, I might come here as well in the long run. Besides that I do not plan to work on new features here atm, so this change of the docs might be on master for a while.

But just for a docs update I also would not mind so much.

mateuszbaran (approved):
LGTM

